PR #57, opened 2019-06-24
Add local overrides for xdebug for php based containers

merk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no     
| Deprecations? | no
| Tests pass?   | yes    
| Fixed tickets | no

Adds overrides for development php.ini for all php based containers instead of just API

### DIFF
--- a/packages/EasyDocker/templates/docker-compose.local.yml.twig
+++ b/packages/EasyDocker/templates/docker-compose.local.yml.twig
@@ -6,9 +6,24 @@ services:
       - ./:/var/www:delegated
       - ./docker/api/development/php.ini:/usr/local/etc/php/conf.d/99-overrides.ini
 
+  artisan:
+    volumes:
+      - ./:/var/www:delegated
+      - ./docker/api/development/php.ini:/usr/local/etc/php/conf.d/99-overrides.ini
+
+  composer:
+    volumes:
+      - ./:/var/www:delegated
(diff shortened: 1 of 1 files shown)


packages/EasyDocker/templates/docker-compose.local.yml.twig
```diff
@@ -6,10 +6,25 @@ services:
...
 
+  artisan:
+    volumes:
+      - ./:/var/www:delegated
+      - ./docker/api/development/php.ini:/usr/local/etc/php/conf.d/99-overrides.ini
+
+  composer:
+    volumes:
+      - ./:/var/www:delegated
+      - ./docker/api/development/php.ini:/usr/local/etc/php/conf.d/99-overrides.ini
```
bradmathews: Will this enable xdebug for composer?  I don't know if that's desired?

merk:
True. Probably better to have a dedicated ini for composer?

bradmathews:
Agree if we need overrides for composer - that's the way to do it.  Is
there currently a use-case for overrides for composer? memory limit?

*Brad Mathews*
Tech Lead

*E*  brad@loyaltycorp.com.au
*T*  1300 134 418
*M*  0414 384 235
Level 1, 1183 Toorak Rd, Camberwell VIC 3124

PR #68, opened 2019-07-09
[EasyDocker] Use .easy directory for easy-docker generated parameter & manifest files

MitchellMacpherson:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

By default, a `.easy` directory will be assumed for parameter and manifest files generated by this tool
Backwards compatibility is present, this will only affect new projects using the tool, when a manifest/params file does not already exist


merk:
❤️

PR #69, opened 2019-07-09
[EasyCfhighlander] Use .easy directory for easy-docker generated parameter & manifest files

MitchellMacpherson:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

Same as https://github.com/loyaltycorp/easy-monorepo/pull/68 but for EasyCfhighlander

> By default, a `.easy` directory will be assumed for parameter and manifest files generated by this tool
> Backwards compatibility is present, this will only affect new projects using the tool, when a manifest/params file does not already exist

### DIFF
--- a/packages/EasyCfhighlander/src/Console/Commands/AbstractTemplatesCommand.php
+++ b/packages/EasyCfhighlander/src/Console/Commands/AbstractTemplatesCommand.php
@@ -114,6 +114,7 @@ abstract class AbstractTemplatesCommand extends Command
     {
         $style = new SymfonyStyle($input, $output);
         $cwd = $input->getOption('cwd') ?? \getcwd();
+        $easyDirectory = $this->getEasyDirectory($cwd);
 
         foreach ($this->getParamResolvers($style) as $param => $resolver) {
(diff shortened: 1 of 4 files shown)


packages/EasyCfhighlander/src/Console/Commands/AbstractTemplatesCommand.php
```diff
@@ -377,6 +383,24 @@ protected function getTemplateName(string $template): string
...
+    /**
+     * Determine if the .easy directory exists
+     *
+     * @param string $cwd
+     *
+     * @return string
+     */
+    private function getEasyDirectory(string $cwd): string
+    {
+        // If easy-docker* file already exists in cwd, .easy directory will not be used/created
```
natepage: Please make sure it looks for the parameters for cfhighlander 😛

MitchellMacpherson: Fixed

PR #71, opened 2019-07-09
[EasyCfhighlander] Added comments to template cfhighlander files to explain file purpose

MitchellMacpherson:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/packages/EasyCfhighlander/templates/code/project.cfhighlander.rb.twig
+++ b/packages/EasyCfhighlander/templates/code/project.cfhighlander.rb.twig
@@ -1,3 +1,5 @@
+# Template for the application, used to produce, validate and publish CloudFormation templates from defined components
+
 CfhighlanderComponent do
 
     Parameters do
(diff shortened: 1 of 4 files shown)


merk:
I'd like to see comments more suited towards non technical (or at least, junior developers) so that we can more easily understand the purposes of these files.

If they're in the repository rather than being abstracted away it means we may need to customise them and without the domain knowledge of whoever built these files we have no chance of understanding.

packages/EasyCfhighlander/templates/code/project.cfhighlander.rb.twig
```diff
@@ -1,3 +1,4 @@
+# AWS ECS Task to update api, worker and cron
```
natepage: Sorry I didn't see we've written that yesterday... But those components are responsible for generating the CloudFormation templates for the different services: api, worker and cron.

Then each of those CloudFormation templates will create ECS tasks resources yes :)

PR #108, opened 2019-12-16
Add a watchlist to EntityChangeSubscriber

merk:
Adds a way to tell the EntityChangeSusbcriber which classes it should emit events for.

If the array is present (not null) the dispatched events will only contain details about the classes that are watched, otherwise no events will be emitted.


natepage:
Implemented in #109

PR #110, opened 2019-12-27
PYMT-1631 BC BREAK Entity Change now tracks changed properties

merk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes 
| BC breaks?    | yes   
| Deprecations? | no 
| Tests pass?   | yes

This PR introduces a major BC break to the EntityChangeSubscriber and will need a major bump.

The goals:
- Instead of passing around opaque arrays of changes, we now pass DTOs that represent each change
- Reduce the complexity of deletion handling - we no longer handle collection deletions (calling ->clear() on a collection) and we never supported reacting to orphan removals, it isnt possible to get them.
- We now add an array of changed properties into the DTOs so that listeners to the change event can decide if they care about the update.
- Changed the way that "delete event enrichment occurs" to reduce complexity in handling edge cases, we now expect the application to implement an interface `DeletedEntityEnrichmentInterface ` to handle adding required deletion information. For example, search will implement the service that adds a search id key to the metadata so it can know what to remove from the indices.


packages/EasyEntityChange/src/Doctrine/EntityChangeSubscriber.php
```diff
@@ -6,41 +6,41 @@
...
+use Doctrine\ORM\UnitOfWork;
+use EonX\EasyEntityChange\DataTransferObjects\DeletedEntity;
+use EonX\EasyEntityChange\DataTransferObjects\UpdatedEntity;
 use EonX\EasyEntityChange\Events\EntityChangeEvent;
-use EonX\EasyEntityChange\Events\EntityDeleteDataEvent;
-use EonX\EasyEntityChange\Exceptions\InvalidDispatcherException;
+use EonX\EasyEntityChange\Interfaces\DeletedEntityEnrichmentInterface;
+use Psr\EventDispatcher\EventDispatcherInterface;
 
 final class EntityChangeSubscriber implements EventSubscriber
```
merk: Make sure this file gets expanded, thx github for hiding it

packages/EasyEntityChange/standards.cfg
```diff
@@ -0,0 +1,2 @@
+# Base application, ensure code coverage is 100%
+PHPUNIT_COVERAGE_MINIMUM_LEVEL=100
```
edwardmurrell-loyaltycorp: :D

edwardmurrell-loyaltycorp:
Where is `DeleteDataEnrichmentInterface` ? I can't find it anywhere

merk:
typoed, it is `DeletedEntityEnrichmentInterface`

packages/EasyEntityChange/composer.json
```diff
@@ -4,11 +4,21 @@
     "type": "library",
     "license": "MIT",
     "require": {
-        "php": "^7.1",
-        "eonx-com/externals": "^1.0"
+        "php": "^7.3",
+        "eonx-com/standards": "^0.3.0",
+        "phpmd/phpmd": "^2.8",
```
natepage: Why are those dependencies in `require` and not `require-dev`?

packages/EasyEntityChange/src/DataTransferObjects/ChangedEntity.php
```diff
@@ -0,0 +1,59 @@
...
+namespace EonX\EasyEntityChange\DataTransferObjects;
+
+/**
+ * This DTO represents a single entity which has changed. Subclasses define the operation
+ * that occurred on the entity.
+ */
+abstract class ChangedEntity
+{
+    /**
+     * @phpstan-var class-string
```
natepage: Why do we need PHPStan annotations? Especially on things that are already defined by their PHPDoc or type hinted by the code.

merk: Because class-string is not string - it explictly limits the string inputs to be instances of classes (which is what we want here)

packages/EasyEntityChange/composer.json
merk: Reason: Derping

packages/EasyEntityChange/src/DataTransferObjects/ChangedEntity.php
merk: I would just annotate the @var as `class-string` but it will break phpstorm. The solution the phpstan/psalm communities have come up with is a prefixed annotation until IDEs support more advanced types

PR #145, opened 2020-03-03
[EasyApiToken] Allow caching for JWKs retrieval

edwardmurrell-loyaltycorp:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | SUBS-160

This enables caching of the JWKS file when using Auth0 when using RS256 signing algorithms


packages/EasyApiToken/composer.json
```diff
@@ -15,7 +15,9 @@
         "auth0/auth0-php": "^5.4",
         "psr/container": "^1.0",
         "laravel/lumen-framework": "^5.5",
-        "symfony/symfony": "^4.2"
```
natepage: Not all projects are using symfony 5 yet, so instead we should set the version to `^4.2 || ^5.0`. And no need to require individual component it is only for dev.

packages/EasyApiToken/src/Factories/Decoders/AbstractJwtTokenDecoderFactory.php
```diff
@@ -96,12 +97,19 @@ protected function createJwtDriver(string $driver, array $options): JwtDriverInt
      */
     private function createAuth0Driver(array $options): Auth0JwtDriver
     {
+        $cache = null;
+        if (array_key_exists('cache_path', $options) && empty($options['cache_path']) === false) {
```
natepage: This can be written as `if (empty($options['cache_path'] ?? null)) === false) {`.

packages/EasyApiToken/src/Factories/Decoders/AbstractJwtTokenDecoderFactory.php
```diff
@@ -96,12 +97,19 @@ protected function createJwtDriver(string $driver, array $options): JwtDriverInt
...
+            $cache = new FileSystemCacheHandler($options['cache_path']);
+        }
+
+
```
natepage: Please remove any unnecessary blank line.

packages/EasyApiToken/src/Factories/Decoders/AbstractJwtTokenDecoderFactory.php
edwardmurrell-loyaltycorp: fixed

packages/EasyApiToken/composer.json
edwardmurrell-loyaltycorp: Must have been borked during merge. Fixed.

PR #162, opened 2020-03-13
[EasyPagination] Add eloquent length aware paginator

cainseing:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | #...   

Added a new paginator for illuminate eloquent


packages/EasyPagination/src/Paginators/EloquentLengthAwarePaginator.php
```diff
@@ -0,0 +1,113 @@
...
+        parent::__construct($startSizeData);
+    }
+
+    public function getTotalItems(): int
+    {
+        if ($this->count !== null) {
+            return $this->count;
+        }
+
+        return $this->createQueryBuilder()->count(\implode(',', $this->getSelect()));
```
ttam: This should also set the count variable for caching

PR #205, opened 2020-04-15
[EasyStandard] Add AvoidPrivateConstructorSniff

ttam:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes

### DIFF
--- a/packages/EasyStandard/src/Sniffs/Classes/AvoidPrivateConstructorSniff.php
+++ b/packages/EasyStandard/src/Sniffs/Classes/AvoidPrivateConstructorSniff.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyStandard\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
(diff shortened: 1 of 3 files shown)


packages/EasyStandard/src/Sniffs/Classes/AvoidPrivateConstructorSniff.php
```diff
@@ -0,0 +1,33 @@
...
+declare(strict_types=1);
+
+namespace EonX\EasyStandard\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+final class AvoidPrivateConstructorSniff implements Sniff
+{
+    public function process(File $phpcsFile, $stackPtr)
```
natepage, line 12: - PHPStan will complain about `$stackPtr` not being defined and we can't actually type hint it so the solution here is to add a PHPDoc bloc for it.
- Also we can add the return type hint `: void`

PR #210, opened 2020-04-20
[EasySecurity/EasyCore] Fixed Registration for argument $targets for PermissionExpressionFunctionProvider and registered ProcessWithLockMiddleware to container 

buddyshontogan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A

- [EasySecurity] Fixed RegisterPermissionExpressionFunctionPass to set targets for PermissionExpressionFunctionProvider.
- [EasyCore] Updated services.yaml to register ProcessWithLockMiddleware.

### DIFF
--- a/packages/EasyCore/src/Bridge/Symfony/Resources/config/services.yaml
+++ b/packages/EasyCore/src/Bridge/Symfony/Resources/config/services.yaml
@@ -6,3 +6,6 @@ services:
 
     EonX\EasyCore\Lock\LockServiceInterface:
         class: EonX\EasyCore\Lock\LockService
+
+    EonX\EasyCore\Bridge\Symfony\Messenger\ProcessWithLockMiddleware:
+        class: EonX\EasyCore\Bridge\Symfony\Messenger\ProcessWithLockMiddleware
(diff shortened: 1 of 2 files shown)


packages/EasyCore/src/Bridge/Symfony/Resources/config/services.yaml
```diff
@@ -6,3 +6,6 @@ services:
 
     EonX\EasyCore\Lock\LockServiceInterface:
         class: EonX\EasyCore\Lock\LockService
+
+    EonX\EasyCore\Bridge\Symfony\Messenger\ProcessWithLockMiddleware:
```
natepage, line 10: Instead of having to type the service class twice, because the service id is basically the service class you can do that:

```yaml
EonX\EasyCore\Bridge\Symfony\Messenger\ProcessWithLockMiddleware: null
```

buddyshontogan: Noted.

PR #218, opened 2020-05-05
[EasyStandard] Added new rectors and sniffs

BOB41K1987:
- Add new rules for quality tools

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes  
| Fixed tickets |


packages/EasyStandard/tests/Sniffs/ControlStructures/ArrangeActAssertSniffTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+use EonX\EasyStandard\Sniffs\ControlStructures\ArrangeActAssertSniff;
+use Symplify\EasyCodingStandardTester\Testing\AbstractCheckerTestCase;
+
+/**
+ * @covers \EonX\EasyStandard\Sniffs\ControlStructures\ArrangeActAssertSniff
+ *
+ * @internal
+ */
+final class ArrangeActAssertSniffTest extends AbstractCheckerTestCase
```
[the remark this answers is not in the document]
BOB41K1987: We can inline `$privatePropertyValue` - this is the easiest solution by far.

packages/EasyStandard/src/Sniffs/Exceptions/ThrowExceptionMessageSniff.php
```diff
@@ -0,0 +1,64 @@
...
+
+        if ($messageToken['code'] === \T_VARIABLE) {
+            return;
+        }
+
+        if ($messageToken['code'] === \T_CLOSE_PARENTHESIS) {
+            return;
+        }
+
+        if (Strings::startsWith($messageToken['content'], "'exceptions.") === false) {
```
natepage, line 44: Hmm.. Ok I think that this sniff is really opinionated but I can see some good in it 😄 

My only real issue is that all translations are not necessarily gonna start with `exceptions.` that depends on the app. I think that this sniff should accept an optional array of prefixes to test against, if the `$messageToken['content']` doesn't start with at least one of them then the sniff should fail.

You can default this list of prefixes to `['exceptions.']` if you want I don't mind.

BOB41K1987: Done.

TomasVotruba:
Hi,

I got a tip for this rule form @natepage 

I really like it :+1: 

@roman-eonx  Would you like to add to @rectorphp core? :)

roman-eonx:
Hi @TomasVotruba,

Which rules are you asking about? Because we have multiple ones here.

TomasVotruba:
Oh, I missed that, sorry.

This `StrictInArrayRector`

TomasVotruba:
I checked the rest and also `AddSeeAnnotationRector` seem useful for common developers. I assume its useful in long test to make it easier to jump between test method and the data it tests.

What do you think?

roman-eonx:
@TomasVotruba yes, we could add those ones, as they can be helpful for other developers and are not specific to our projects.

TomasVotruba:
Thank you

PR #246, opened 2020-06-04
[EasyDecision] Implement decision types mapping

BOB41K1987:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


packages/EasyDecision/src/Bridge/Laravel/EasyDecisionServiceProvider.php
```diff
@@ -70,6 +80,10 @@ function (): ExpressionLanguageFactoryInterface {
             }
         );
 
+        $this->app->singleton(MappingProviderInterface::class, static function (): MappingProviderInterface {
```
roman-eonx, line 83: Why do we need to duplicate it here? We've registered this interface in the `register` method already.

packages/EasyDecision/src/Interfaces/MappingProviderInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDecision\Interfaces;
+
+interface MappingProviderInterface
+{
+    /**
+     * Returns decision type.
```
roman-eonx, line 10: Please remove this PHPDoc, as in this package it's preferred to add them only when some annotations are needed.

packages/EasyDecision/src/Providers/AbstractMappingProvider.php
```diff
@@ -0,0 +1,28 @@
...
+
+abstract class AbstractMappingProvider implements MappingProviderInterface
+{
+    /**
+     * @var string[]
+     */
+    protected $typesMapping = [];
+
+    /**
+     * {@inheritdoc}
```
roman-eonx, line 18: Please remove this.

packages/EasyDecision/src/Providers/ConfigMappingProvider.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDecision\Providers;
+
+final class ConfigMappingProvider extends AbstractMappingProvider
+{
+    /**
+     * ConfigMappingProvider constructor.
```
roman-eonx, line 10: Please remove the description. Leave only `@param`.

packages/EasyDecision/src/Providers/ConfigMappingProvider.php
```diff
@@ -0,0 +1,32 @@
...
+     *
+     * @param string[] $decisionsConfig
+     */
+    public function __construct(array $decisionsConfig)
+    {
+        $this->setTypesMapping($decisionsConfig);
+    }
+
+    /**
+     * Sets types mapping.
```
roman-eonx, line 20: The same.

packages/EasyDecision/tests/Bridge/Symfony/EasyDecisionBundleTest.php
```diff
@@ -118,6 +120,47 @@ static function (DecisionInterface $decision): void {
         ];
     }
 
+    /**
+     * Tests trying to use configuration with nonexistent decision class throws InvalidConfigurationException.
```
roman-eonx, line 124: Please remove the description. Here and below.

packages/EasyDecision/src/Interfaces/MappingProviderInterface.php
BOB41K1987: Done.

packages/EasyDecision/src/Providers/AbstractMappingProvider.php
BOB41K1987: Done.

packages/EasyDecision/src/Providers/ConfigMappingProvider.php
BOB41K1987: Done.

packages/EasyDecision/src/Providers/AbstractMappingProvider.php
```diff
@@ -0,0 +1,25 @@
...
+use EonX\EasyDecision\Interfaces\MappingProviderInterface;
+
+abstract class AbstractMappingProvider implements MappingProviderInterface
+{
+    /**
+     * @var string[]
+     */
+    protected $typesMapping = [];
+
+    public function getDecisionType(string $name): string
```
natepage, line 17: Here I would like us to have an "early return". And we should also validate that the value is a valid classname.

packages/EasyDecision/src/Providers/ConfigMappingProvider.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public function __construct(array $decisionsConfig)
+    {
+        $this->setTypesMapping($decisionsConfig);
+    }
+
+    /**
+     * @param string[] $decisionsConfig
+     */
+    private function setTypesMapping(array $decisionsConfig): void
```
natepage, line 20: I think that this method should be on the abstract mapping provider, this way regardless of where the mapping is coming from the abstract class provides the logic to validate the entries.

packages/EasyDecision/src/Providers/AbstractMappingProvider.php
natepage: Please make sure that we have 2 different exception messages for each use case to make DX better:

- "Decision for name ... not configured"
- "Decision for name ... not a valid classname"

packages/EasyDecision/src/Providers/ConfigMappingProvider.php
BOB41K1987: Done.

packages/EasyDecision/src/Providers/AbstractMappingProvider.php
BOB41K1987: I have added check for valid classname and made a separate test for that. I'm not sure how we can use "early return" here though.

packages/EasyDecision/src/Decisions/DecisionFactory.php
```diff
@@ -201,4 +206,11 @@ private function instantiateDecision(string $decisionType): DecisionInterface
             DecisionInterface::class
         ));
     }
+
+    public function createByName(string $name): DecisionInterface
```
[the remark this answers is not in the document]
BOB41K1987: Ok, done.

packages/EasyDecision/src/Decisions/DecisionFactory.php
```diff
@@ -201,4 +209,11 @@ private function instantiateDecision(string $decisionType): DecisionInterface
...
+    {
+        $decision = $this->mappingProvider->getDecisionType($name);
+
+        return $this->configureDecision(new $decision());
```
natepage, line 217: Can you please pass the `$name` as a constructor parameter to the new `$decision` and we're all good to go! 😄

PR #251, opened 2020-06-16
[EasyRandom] Register UuidV4 generators

albertlabarentojr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes


packages/EasyRandom/readme.md
```diff
@@ -71,6 +71,16 @@ $randomGenerator = (new \EonX\EasyRandom\RandomGenerator())->setUuidV4Generator(
 $uuidV4 = $randomGenerator->uuidV4();
 ```
 
+#### In Symfony configure your preferred UUID V4 generator
```
natepage, line 74: Here can you please simply explain that this is possible because by default the bundle registers the built-in generators as services.

packages/EasyRandom/src/Bridge/Symfony/Resources/config/services.yaml
```diff
@@ -3,5 +3,14 @@ services:
         autowire: true
```
natepage, line 3: Can you also add `private: true` so this way we're sure that the services are private by default so they will be removed from the container if not used.

packages/EasyRandom/readme.md
```diff
@@ -71,6 +71,18 @@ $randomGenerator = (new \EonX\EasyRandom\RandomGenerator())->setUuidV4Generator(
...
+
+To make this possible, the bundle will register the built-in generators as services.
+
+- symfony_uuid4
```
natepage, line 78: I think here we should put the full alias so it makes sense to the readers why in the following example we're using `easy_random.symfony_uuid4`

PR #278, opened 2020-07-15
[EasyCore] Add DateTimeMicrosecondsType support for PostgreSQL

BOB41K1987:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


packages/EasyCore/src/Doctrine/Type/DateTimeMicrosecondsType.php
```diff
@@ -27,6 +28,11 @@ final class DateTimeMicrosecondsType extends Type
      */
     public const FORMAT_DB_TIMESTAMP = 'TIMESTAMP';
 
+    /**
+     * @var string
+     */
+    public const FORMAT_DB_TIMESTAMP_WO_TIMEZONE = 'TIMESTAMP(6) WITHOUT TIME ZONE';
```
natepage, line 34: Because we're storing every datetime in UTC, why `without time zone`?

Issue #268, opened 2020-07-08
Symfony PHP Services Config


natepage:
In preparation of Symfony 6, an initiative made by the core team was to move XML and YAML services config to PHP (https://github.com/symfony/symfony/issues/37186)

Let's do the same and get ready for the furure! 😄 

This will require us to:

- Change the DI file loader to use PHPFileLoader
- Replace all `Bridge/Symfony/Resources/config/*.yaml` and `*.xml` files to `*.php`

Here is an example: https://github.com/symfony/symfony/commit/18fcb5f956779cf6faaafe5d2087c80005182202

BOB41K1987:
Here is a list of config files that need to be changed:

- [ ] packages/EasyApiToken/src/Bridge/Symfony/Resources/config/services.xml
- [ ] packages/EasyAsync/src/Bridge/Symfony/Resources/config/implementations/doctrine.yaml
- [ ] packages/EasyAsync/src/Bridge/Symfony/Resources/config/services.yaml
- [ ] packages/EasyAwsCredentialsFinder/src/Bridge/Symfony/Resources/config/services.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/debug.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/iri_converter.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/maker_commands.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/no_properties_api_resource.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/pagination.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/simple_data_persister.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/easy_async_listeners.yaml
- [ ] packages/EasyCore/src/Bridge/Symfony/Resources/config/services.yaml
- [ ] packages/EasyDecision/src/Bridge/Symfony/Resources/config/services.yaml
- [ ] packages/EasyDecision/src/Bridge/Symfony/Resources/config/use_expression_language.yaml
- [ ] packages/EasyPagination/src/Bridge/Symfony/Resources/config/services.yaml
- [ ] packages/EasyPsr7Factory/src/Bridge/Symfony/Resources/config/services.xml
[...]

PR #294, opened 2020-08-03
Migrate Symfony Services Config to PHP

BOB41K1987:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #268


packages/EasyAsync/src/Bridge/Symfony/Resources/config/implementations/doctrine.php
```diff
@@ -0,0 +1,26 @@
...
+    $services = $containerConfigurator->services();
+
+    $services->defaults()
+        ->autowire()
+        ->autoconfigure();
+
+    $services->set(DataCleanerInterface::class, DataCleaner::class);
+
+    $services->set(JobLogPersisterInterface::class, JobLogPersister::class)
+        ->bind('$table','%easy_async_job_logs_table%');
```
natepage, line 22: Any particular reason why we're using `bind` instead of `arg`?

natepage: Ok it was like that in the YAML file :D nvm

natepage:
@BOB41K1987 Nice work! I'm really sorry but I didn't realise that you've based your work on `3.x`, I've created that branch one night as a starting point but it is not ready yet for us to start working on it.

And also any BC compatible changes we make to prepare for v3 can be done in master directly as it shouldn't break anything. Can you please base the update from master, thank you.

natepage:
Done in #296.

packages/EasyAsync/src/Bridge/Symfony/Resources/config/implementations/doctrine.php
TomasVotruba: Good question, `bind()` is confusing here for me too. I'm curious, what's the different in 
behavior?

I saw bind() only in defaults() definition above.

natepage: End result is the same, at least for us 😄 Usually I use `arg` for a specific service but it might be one of those packages I've written overnight because of an idea I had, and I was young at the time! At least younger than now anyway... 😉

PR #295, opened 2020-08-04
[EasyErrorHandler] Added type casting in config

skrut:
- added type casting in config (allows to pass the env variable value from `phpunit.xml`)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets |

### DIFF
--- a/packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
+++ b/packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
@@ -190,7 +190,9 @@ class Handler implements ExceptionHandler
      */
     protected function isExtendedResponse(): bool
     {
-        return (bool)$this->config->get('easy-error-handler.use_extended_response', false) === true;
+        $extendedResponse = (bool)$this->config->get('easy-error-handler.use_extended_response', false);
+
+        return $extendedResponse === true;
     }
 
     /**
--- a/packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
+++ b/packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
@@ -190,10 +190,7 @@ class Handler implements ExceptionHandler
      */
     protected function isExtendedResponse(): bool
     {
-        $useExtendedResponse = \filter_var(
-            $this->config->get('easy-error-handler.use_extended_response', false),
-            \FILTER_VALIDATE_BOOLEAN
-        );
+        $useExtendedResponse = (bool)$this->config->get('easy-error-handler.use_extended_response', false);
 
         return $useExtendedResponse === true;
     }
--- a/packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
+++ b/packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
@@ -190,9 +190,7 @@ class Handler implements ExceptionHandler
      */
     protected function isExtendedResponse(): bool
     {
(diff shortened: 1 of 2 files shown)


packages/EasyErrorHandler/src/Bridge/Laravel/config/easy-error-handler.php
```diff
@@ -7,7 +7,10 @@
     /**
      * Use extended error response with exception message, trace, etc.
      */
-    'use_extended_response' => \env('EASY_ERROR_HANDLER_USE_EXTENDED_RESPONSE', false),
+    'use_extended_response' => \filter_var(
```
egor-dev, line 10: Please, cast to boolean in the `isExtendedResponse` method instead. This package uses this approach.

skrut: Done.

packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
```diff
@@ -190,7 +190,12 @@ protected function getResponseStatusCode(Exception $exception): int
      */
     protected function isExtendedResponse(): bool
     {
-        return $this->config->get('easy-error-handler.use_extended_response', false) === true;
+        $useExtendedResponse = \filter_var(
+            $this->config->get('easy-error-handler.use_extended_response', false),
+            \FILTER_VALIDATE_BOOLEAN
+        );
+
+        return $useExtendedResponse === true;
```
egor-dev, line 198: No need for `filter_var` just `(bool)`.

skrut: Done.

egor-dev: Sorry, please, revert to your approach. Because in my way `"false"` will cast to `true`.

skrut: Done.

packages/EasyErrorHandler/src/Bridge/Laravel/config/easy-error-handler.php
natepage: I don't understand why do we have to do all that? The laravel `env()` method is already taking care of casting boolean values. Isn't that working?

```php
// vendor/illuminate/support/Env.php

switch (strtolower($value)) {
	case 'true':
	case '(true)':
		return true;
	case 'false':
	case '(false)':
		return false;
	case 'empty':
	case '(empty)':
		return '';
	case 'null':
	case '(null)':
		return;
}
```

packages/EasyErrorHandler/src/Bridge/Laravel/Handler/Handler.php
```diff
@@ -190,7 +190,7 @@ protected function getResponseStatusCode(Exception $exception): int
      */
     protected function isExtendedResponse(): bool
     {
-        return $this->config->get('easy-error-handler.use_extended_response', false) === true;
+        return (bool)$this->config->get('easy-error-handler.use_extended_response', false) === true;
```
natepage, line 193: I'm always confuse with casting, does it apply only to the left side of the expression?

If not, we might have some surprises. Maybe we should be cautious and write it this way:

```php
$responseExtended = (bool)$this->config->get('easy-error-handler.use_extended_response', false);

return $responseExtended === true;
```

skrut: Done.

Issue #297, opened 2020-08-05
[EasyPagination] Doctrine ORM Paginator incorrect total items


jordancachero:
**Describe the bug**
The query that counts total items does not remove duplicates (w/c doctrine does), if JOIN is used.
The query that counts total items becomes incorrect when Count is used with groupBy.

**To Reproduce**
On a repository that uses DoctrineOrmLengthAwarePaginator::setCriteria, add a join that will duplicate the row. Doctrine ORM hydrates and remove duplicates, but the counting does not.
Also add a groupBy clause and this will cause the counting to be incorrect.

**Expected behavior**
Total Items count should be correct.

natepage:
@jordancachero Can you please provide an example of the code using the paginator with the issue. Would be helpful to understand what's going on, cheers.

roman-eonx:
@jordancachero I've changed the issue title by adding the package name as a prefix. We usually do this to always see what package is affected by the issue. Btw, the same goes for pull requests in monorepo :wink:

jordancachero:
Issue for when `Total Items Count > Actual Items`
- This is caused by Joins w/c creates duplicate rows

```
return $this->createLengthAwarePaginator($this->getEntityClass(), 'ga')
            ->setCriteria(static function (QueryBuilder $queryBuilder) use ($awardFilterData): void {
                     if ($awardFilterData->getNomineeId() !== null) {
                          $queryBuilder->innerJoin('ga.awardNominations', 'an')
                               ->andWhere('an.nominee = :nomineeId')
                               ->setParameter('nomineeId', $awardFilterData->getNomineeId())
                    }
....
```

Issue for when `Actual Items > Total Items Count`
- This is caused when group by is used. Ex scenario, is listing the Awards, sorted by nominations made count.

```
return $this->createLengthAwarePaginator($this->getEntityClass(), 'ga')
            ->setCriteria(static function (QueryBuilder $queryBuilder) use ($awardFilterData): void {
....
                       $queryBuilder
                            ->leftJoin('ga.awardNominations', 'an')
                            ->leftJoin('an.period', 'gap', Join::WITH, 'gap.currentPeriod = :currentPeriod')
                            ->setParameter('currentPeriod', true)
                            ->orderBy('COUNT(an.awardId)', $order)
                            ->groupBy('ga.awardId');
....
```

natepage:
@jordancachero @roman-eonx 

[...]

PR #298, opened 2020-08-05
[EasyPagination] Fix total items count for doctrine orm

jordancachero:
- Fix for doctrine orm get total items.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no 
| Deprecations? | no
| Tests pass?   | yes 
| Fixed tickets | https://github.com/eonx-com/easy-monorepo/issues/297


packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
```diff
@@ -43,7 +43,16 @@ public function getTotalItems(): int
...
+
+        $queryBuilder = $this->createQueryBuilder()->select(
+            \sprintf(
+                'COUNT(DISTINCT %s) as %s', // Need to distinct to remove duplicates caused by joins.
+                $this->fromAlias,
+                $countAlias
+            )
+        );
+
+        $queryBuilder->resetDQLPart('groupBy'); // Reset groupBy to remove issue when counting.
```
natepage, line 55: What issue is that solving?

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
```diff
@@ -43,7 +43,16 @@ public function getTotalItems(): int
         }
 
         $countAlias = \sprintf('_count%s', $this->fromAlias ? \sprintf('_%s', $this->fromAlias) : '');
-        $queryBuilder = $this->createQueryBuilder()->select(\sprintf('COUNT(1) as %s', $countAlias));
+
+        $queryBuilder = $this->createQueryBuilder()->select(
+            \sprintf(
+                'COUNT(DISTINCT %s) as %s', // Need to distinct to remove duplicates caused by joins.
+                $this->fromAlias,
```
natepage, line 50: `fromAlias` is nullable, what happen when it is not set? Should we default it to something?

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
jordancachero: If a group by exist this will cause the result to have multiple rows.

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
jordancachero: okay, gonna test what should be the proper default if fromAlias is null.

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
natepage: Ok but here we're modifying the SQL query of the app using the package without telling them. What if they actually need the group by?

If it's a real issue to have a groupBy here, I think we should throw an exception to let the app know that they can't add a groupBy in the criteria to count items.

And what happen when we remove the groupBy to count the items but not to fetch them? Doesn't that create inconsistencies as well?

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
natepage: Keep me posted 😉

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
jordancachero: Example scenario for when groupBy is needed, is for feature that allows sorting a list of Awards by count of Nominations, so restricting usage of groupBy here will not allow that scenario.

As per testing, items fetched is hydrated, so the itemsPerPage still is per hydrated object count(even with group by). But with totalItems counting, if groupBy exist, this will cause the result to multiple rows and causes inconsistent total count. Removing groupBy + with Count distinct when counting total items seems to be correct way on counting total items.

Haven't tested all scenarios.

Maybe we should move this to a branch, and use it on RR to be safe first?

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
jordancachero: Having hard time with this, null alias is only when this trait is used on DoctrineDbalLengthAwarePaginator. Checking EasyRepository package only have AbstractPaginatedDoctrineOrmRepository, that only DoctrineOrmLengthAwarePaginator.

natepage: `eonx-com/easy-async` is using the dbal paginator, you can test it there.

packages/EasyPagination/src/Traits/DoctrinePaginatorTrait.php
jordancachero: Was thinking this may affect DoctrineDbalLengthAwarePaginator, so maybe move the `getTotalItems` on each subclass instead of in this trait.

DoctrineDbalLengthAwarePaginator contain the old getTotalItems
DoctrineOrmLengthAwarePaginator contain the updated getTotalItems w/c will solve the Alias issue as well.

natepage: I don't think that this a good approach because the intent was to keep a consistent logic across multiple platforms but now we have 2 different ways of counting items. Have a look at my comment in the issue 😉

jordancachero: Yes but DoctrinePaginatorTrait already contains an abstract method: `doGetTotalItems` w/c causes paginators to have separate logic when getting total items.

Added comment on issue: https://github.com/eonx-com/easy-monorepo/issues/297#issuecomment-669573871

Issue #187, opened 2020-04-01
[EasyDecision] Symfony WebProfiler integration


natepage:
Create an integration for the Symfony WebProfiler to be able to see decision data in the profiler to make debugging easier.

natepage:
What should we profile?

albertlabarentojr:
Issue taken

albertlabarentojr:
Hi @natepage  , I'm new to this web profiler integration, however, I found something interesting
https://symfony.com/doc/current/profiler.html#data-collectors

Maybe this can be a good start, just an idea

natepage:
@albertlabarentojr Yes that's it 👍 

For examples about existing DataCollectors you can have a look at:

- https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasySecurity/src/Bridge/Symfony/DataCollector/SecurityContextDataCollector.php
- https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasyWebhook/src/Bridge/Symfony/DataCollector/WebhookDataCollector.php

The way to register them in the container:

- https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasySecurity/src/Bridge/Symfony/Resources/config/services.php#L50
- https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasyWebhook/src/Bridge/Symfony/Resources/config/debug.php#L21

And the associated twig templates (don't forget the icon 😉 ):

- https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasySecurity/src/Bridge/Symfony/Resources/views/Collector/security_context_collector.html.twig
- https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasyWebhook/src/Bridge/Symfony/Resources/views/Collector/webhook_collector.html.twig

albertlabarentojr:
You literally gave me everything I need, thank you @natepage !

natepage:
Done in #317

PR #317, opened 2020-08-15
[EasyDecision] Symfony WebProfiler integration

albertlabarentojr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #187


packages/EasyDecision/src/Bridge/Symfony/DataCollector/DecisionContextDataCollector.php
```diff
@@ -0,0 +1,80 @@
...
+
+use EonX\EasyDecision\Interfaces\DecisionContextInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+
+final class DecisionContextDataCollector extends DataCollector
+{
+    /**
+     * @const string
```
natepage, line 14: I'm not sure that `@const` exist, let's use `@var`.

packages/EasyDecision/src/Bridge/Symfony/DataCollector/DecisionContextDataCollector.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public const NAME = 'easy_decision.decision_context_collector';
+
+    /**
+     * @var \EonX\EasyDecision\Interfaces\DecisionContextInterface
+     */
+    private $decisionContext;
+
+    public function __construct(DecisionContextInterface $decisionContext)
```
natepage, line 23: The `DecisionContextInterface` isn't defined as a service, and the main reason why it's because there is one context for each decision made. What happen when multiple decisions are made during the same request?

packages/EasyDecision/src/Interfaces/DecisionContextInterface.php
```diff
@@ -4,7 +4,7 @@
 
 namespace EonX\EasyDecision\Interfaces;
 
-interface ContextInterface
+interface DecisionContextInterface
```
natepage, line 7: The objective of this PR is to add Symfony Web Profiler support for this package, that shouldn't require any naming changes to the existing classes AND changing the name of an interface like that is a BC break so it should be done either safely by supporting both name until next major version or simply wait for the next major version.

What happen if a project is using this interface as a parameter type hint because they're using the decision context in their code? It will break. Let's revert that and if we really want to change it then it should be its own PR.

packages/EasyDecision/src/Bridge/Symfony/DataCollector/DecisionContextDataCollector.php
albertlabarentojr: Very noted @natepage ! 👍

packages/EasyDecision/src/Decisions/DecisionFactory.php
```diff
@@ -47,6 +53,7 @@ final class DecisionFactory implements DecisionFactoryInterface
      * @param null|iterable<mixed> $configurators
      */
     public function __construct(
+        DecisionAggregatorInterface $decisionAggregator,
```
natepage, line 56: Like I said on Slack, nice try, I've thought about doing the same thing in other packages but if you're adding a brand new service only for Symfony DataCollector then simply just modify the existing service to expose the data 😄 

So I don't think we need this aggregator, simply have a public method to get the list of configured decisions, inject this decision factory in the data collector and that's it!

packages/EasyDecision/src/Decisions/DecisionFactory.php
```diff
@@ -155,6 +170,8 @@ static function (DecisionConfiguratorInterface $first, DecisionConfiguratorInter
             $configurator->configure($decision);
         }
 
+        $this->decisionAggregator->addDecision($decision, $configurators);
```
natepage, line 173: Even after removing the aggregator make sure that you collect the configurators for a decision within the foreach above because as you can see configurator can implement the `RestrictedDecisionConfiguratorInterface` and they might not support the given decision so they shouldn't be in the array for that decision.

natepage (changes requested):

packages/EasyDecision/src/Decisions/DecisionFactory.php
albertlabarentojr: haha this was a mistake, Added tests to make sure it's working as expected

natepage (approved):

PR #340, opened 2020-09-04
Rector, ECS and PHPStan update

TomasVotruba:
## Updates

- Rector 0.7 → 0.8
- Symplify family 7.x → 8.2
- PHPStan 0.11 → 0.12

## Cleanup

- move Sniff Fixtures closer to Fixture tests, to prevent jumping from file and fixture and crazy super long fixture paths - this is standard for both Rector and ECS
- remove obvious docblock comments
- monorepo-builder.yml to monorepo-builder.php
- changelog-linker.yml to changelog-linker.php
- ecs.yml to ecs.php

To read more about YAML to PHP, see:

- [10 Cool Features You Get after switching from YAML to PHP Configs](https://tomasvotruba.com/blog/2020/07/16/10-cool-features-you-get-after-switching-from-yaml-to-php-configs/)

- [How to Switch from YAML/XML Configs to PHP Today with Migrify](https://tomasvotruba.com/blog/2020/07/27/how-to-switch-from-yaml-xml-configs-to-php-today-with-migrify/)

PR #344, opened 2020-09-08
[ECS] Extend ruleset with array list and line-lenght rules

TomasVotruba:
Follow up to https://github.com/eonx-com/easy-monorepo/pull/340

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -163,7 +163,7 @@
             "@check-security",
             "@check-tests"
         ],
-        "check-ecs": "vendor/bin/ecs check packages --clear-cache",
+        "check-ecs": "vendor/bin/ecs check --clear-cache",
         "check-monorepo": [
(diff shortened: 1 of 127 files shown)


composer.json
```diff
@@ -163,7 +163,7 @@
             "@check-security",
             "@check-tests"
         ],
-        "check-ecs": "vendor/bin/ecs check packages --clear-cache",
+        "check-ecs": "vendor/bin/ecs check --clear-cache",
```
TomasVotruba, line 166: This argument hear silently ignored the `ecs.php` paths setup. 
Using only `ecs.php` is best practise as it's hard to forget all the paths and there is 1 place to change them

natepage: @TomasVotruba ok thanks for the tip

natepage (approved):

PR #357, opened 2020-09-21
[EasyCore] Disable trim string middleware/denormalizer by default

DKeeper:
- We disable trimming for Symfony by default, so each app should explicitly enable trimming
- Get rid of `enabled` flag for Laravel, because an actual trimming is enabled by the addition/deletion of TrimStrings middleware in an app


egor-dev (approved):

natepage (changes requested):
Ok but are we planning to fix the actual issue?

egor-dev:
> Ok but are we planning to fix the actual issue?

@natepage what do you mean?

natepage:
@egor-dev I mean that there is a problem with the code because it doesn't want in my project so we need to fix it 😄

roman-eonx (approved):

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
```diff
@@ -0,0 +1,200 @@
...
+ * @internal
+ */
+final class TrimStringsDenormalizerTest extends AbstractSymfonyTestCase
+{
+    /**
+     * @return mixed[]
+     *
+     * @see testSupportsDenormalizationReturnsTrue
+     */
+    public function provideCorrectDataForDenormalization(): array
```
egor-dev, line 26: Let's change to `provideDataForSupportsDenormalization` and combine with `provideIneligibleDataForDenormalization`. And use all of the sets in one `testSupportsDenormalizationReturnsExpectedResult`.

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
```diff
@@ -0,0 +1,200 @@
...
+    {
+        return [
+            'context has not `TRIM_STRINGS_ALREADY_CALLED` key' => [
```
egor-dev, line 29: I don't think we need this set. You could just add non-empty context for the other two sets.

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
```diff
@@ -0,0 +1,200 @@
...
+    public function testDenormalizeSucceedsWithTrimValue($data, $expectedResult): void
+    {
+        $type = 'no-matter';
+        $format = null;
+        $context = [];
+        $except = ['some-key'];
+        $expectation = static function (MockInterface $mock): void {
+            $mock->shouldNotReceive('trim');
+        };
+        if (\is_string($data) || \is_array($data)) {
```
egor-dev, line 125: This check is not needed, because the data provider contain only a string and an array.

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
```diff
@@ -0,0 +1,200 @@
...
+            $expectation = static function (MockInterface $mock) use ($data, $except, $expectedResult): void {
+                $mock->shouldReceive('trim')->once()->with($data, $except)->andReturn($expectedResult);
+            };
+        }
+        /** @var \EonX\EasyCore\Helpers\StringsTrimmerInterface $trimmer */
+        $trimmer = $this->mock(StringsTrimmerInterface::class, $expectation);
```
egor-dev, line 131: Why not use real `RecursiveStringsTrimmer` as it does not contain any dependency.

egor-dev (changes requested):

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
DKeeper: Done.

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
DKeeper: Done.

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
DKeeper: Fixed.

packages/EasyCore/tests/Bridge/Symfony/Serializer/TrimStringsDenormalizerTest.php
DKeeper: Done.

egor-dev (dismissed):

egor-dev (approved):

roman-eonx (approved):

Issue #388, opened 2020-10-08
[EasyDecision] AddRestrictedRulesDecisionConfigurator 


albertlabarentojr:
**Is your feature request related to a problem? Please describe.**
When dealing with rules, I have to create a custom DecisionConfigurator just to register all the rules. It can be a pain if we have growing decisions overtime. 

**Describe the solution you'd like**
What if the package provides Rules to implement `RestrictedRuleInterface::supportsDecision(DecisionInterface $decision): bool` and `AddRestrictedRulesDecisionConfigurator` will take care the rest.
Here is the POC as a reference:

https://github.com/eonx-com/gift-card-solutions/blob/hotfix/refactor-easy-decisions-to-use-configurator/src/Externals/EasyDecision/Configurators/AddRestrictedRuleDecisionConfigurator.php

**Additional context**
As per @natepage "however I can see that if a rule isn't a restricted one you don't add it to the decision by default, I think it should be the opposite, if a rule isn't restricted then it means it can go any decision"

natepage:
Done in #389

PR #389, opened 2020-10-08
[EasyDecision] Implemented AddRestrictedRulesDecisionConfigurator

albertlabarentojr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no 
| Tests pass?   | yes
| Fixed tickets | #388


packages/EasyDecision/src/Configurators/AddRestrictedRulesDecisionConfigurator.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        parent::__construct($priority);
+
+        $this->restrictedRules = $restrictedRules;
+    }
+
+    public function configure(DecisionInterface $decision): void
+    {
+        foreach ($this->restrictedRules as $rule) {
+            if (($rule instanceof RestrictedRuleInterface) === false) {
```
natepage, line 30: I think that if a rule isn't restricted then it should be added to any decision by default.

packages/EasyDecision/src/Interfaces/RestrictedRuleInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDecision\Interfaces;
+
+interface RestrictedRuleInterface extends RuleInterface
```
natepage, line 7: Maybe we should create some default abstract restricted rule, because the most common use case to restrict a rule I can think of would be based on its name. So maybe an abstract "name restricted" rule that implements `supportsDecision()` and automatically check the decision name against the one define by the rule, using an abstract protected method, could be cool

natepage (changes requested):

packages/EasyDecision/src/Configurators/AddRestrictedRulesDecisionConfigurator.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        parent::__construct($priority);
+
+        $this->rules = $rules;
+    }
+
+    public function configure(DecisionInterface $decision): void
+    {
+        foreach ($this->rules as $rule) {
+            if (($rule instanceof RestrictedRuleInterface) === false) {
```
natepage, line 30: We could have only 1 if statement here:

```php
if ($rule instanceof RestrictedRuleInterface === false || $rule->supportsDecision($decision)) {
	$decision->addRule($rule);
}
```

natepage (changes requested):
Look nice but I can't see any tests for it

packages/EasyDecision/src/Configurators/AddRestrictedRulesDecisionConfigurator.php
```diff
@@ -0,0 +1,41 @@
...
+                $decision->addRule($rule);
+
+                continue;
+            }
+
+            if ($rule->supportsDecision($decision) === true) {
```
roman-eonx, line 36: I think we should check an instance of `RestrictedRuleInterface` here explicitly. Just in case we add more `RuleInterface` descendants in the future.

packages/EasyDecision/src/Configurators/AddRestrictedRulesDecisionConfigurator.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDecision\Configurators;
+
+use EonX\EasyDecision\Interfaces\DecisionInterface;
+use EonX\EasyDecision\Interfaces\RestrictedRuleInterface;
+
+final class AddRestrictedRulesDecisionConfigurator extends AbstractConfigurator
```
roman-eonx, line 10: As soon as now we process not only restricted rules in this configurator, we should rename it to something more generic. Maybe `AddRulesDecisionConfigurator` or `AddRulesConfigurator`.

roman-eonx (changes requested):

packages/EasyDecision/src/Configurators/AddRulesDecisionConfigurator.php
```diff
@@ -0,0 +1,35 @@
...
+    private $rules;
+
+    /**
+     * @param iterable<\EonX\EasyDecision\Interfaces\RuleInterface> $rules
+     */
+    public function __construct(iterable $rules, ?int $priority = null)
+    {
+        parent::__construct($priority);
+
+        $this->rules = $rules;
```
natepage, line 24: I would like to add some logic to make sure we have only rules in this, have a look at other packages how we filter the iterables

packages/EasyDecision/tests/Stubs/DecisionStub.php
```diff
@@ -0,0 +1,55 @@
...
+     * @var mixed
+     */
+    private $output;
+
+    /**
+     * @var null|\Closure
+     */
+    private $outputClosure;
+
+    public function __construct(
```
natepage, line 26: I can't see anywhere in the test where we're using the other constructor arguments, why having them?

natepage (changes requested):

phpstan.neon
```diff
@@ -102,6 +102,9 @@ parameters:
         - message: '#Cannot call method addFunction\(\) on EonX\\EasyDecision\\Expressions\\Interfaces\\ExpressionLanguageInterface\|null#'
           path: packages/EasyDecision/tests/Bridge/Symfony/Stubs/Configurators/TypeRestrictedExpressionFunctionConfigurator.php
 
+        - message: '#Parameter \#1 \$rules of class EonX\\EasyDecision\\Configurators\\AddRulesDecisionConfigurator constructor expects iterable<EonX\\EasyDecision\\Interfaces\\RuleInterface\>, array<int, class@anonymous/EasyDecision/tests/Configurators/AddRulesDecisionConfiguratorTest\.php\:75\|EonX\\EasyDecision\\Tests\\Stubs\\RestrictedRuleStub\> given#'
```
natepage, line 105: Instead of adding this, you should modify the phpdoc of the configurator to accept mixed, because it's the truth, it can deal with anything thanks to `filterRules()`

natepage (changes requested):

natepage (approved):

PR #405, opened 2020-10-21
static fixes

TomasVotruba:
Fixes ↓ and similar

image

### DIFF
--- a/packages/EasyApiToken/src/Interfaces/ApiTokenDecoderInterface.php
+++ b/packages/EasyApiToken/src/Interfaces/ApiTokenDecoderInterface.php
@@ -23,5 +23,3 @@ interface ApiTokenDecoderInterface
 
     public function getName(): string;
 }
-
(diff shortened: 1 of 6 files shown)


packages/EasyApiToken/src/Interfaces/ApiTokenDecoderInterface.php
```diff
@@ -23,5 +23,3 @@ public function decode($request): ?ApiTokenInterface;
 
     public function getName(): string;
 }
-
-\class_alias(ApiTokenDecoderInterface::class, EasyApiTokenDecoderInterface::class);
```
TomasVotruba, line 27: This easily messes up PHPStan and Rector. Why is this here BTW?

If it's legacy interface, it should be here to inherit from this one

TomasVotruba:
Ready to merge. The CI on `master` should pass now

TomasVotruba:
:+1:

PR #406, opened 2020-10-22
[CI] add all packages to split tests

TomasVotruba:


packages/EasyWebhook/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -46,7 +46,7 @@
     // Webhook Client
     $services
         ->set(WebhookClientInterface::class, WebhookClient::class)
-        ->arg('$configurators', tagged_iterator(BridgeConstantsInterface::TAG_WEBHOOK_CONFIGURATOR))
+        ->arg('$configurators', tagged(BridgeConstantsInterface::TAG_WEBHOOK_CONFIGURATOR))
```
TomasVotruba, line 49: The `tagged_iterator` exists only since Symfony 4.4. So it failed on Symfony 4.2, now minimal version

- The `tagged` works on Symfony 4
- The `tagged_iterator` works on Symfony 4.4 and 5+

https://github.com/symfony/symfony/commit/ab8fb1868f3576b17e4b5a0a57c875cfadbb2bb5#diff-0f4455419d7d5f79423475e597c88c6901a7d00871165ea2967da1183390cf56

We have to use both, depending on Symfony version

natepage:
@TomasVotruba Thank you for the PR, I will close it now and create a new PR for this as we're planning on implementing a dynamic matrix in github actions and we also removed some packages since this PR was opened.

PR #410, opened 2020-10-28
[EasyDecision] Reset decision configurators

elloyd-cruz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | #...


packages/EasyDecision/src/Bridge/Symfony/DependencyInjection/EasyDecisionExtension.php
```diff
@@ -43,5 +46,12 @@ public function load(array $configs, ContainerBuilder $container): void
         $container
             ->autowire(MappingProviderInterface::class, ConfigMappingProvider::class)
             ->setArgument('$decisionsConfig', $config['type_mapping'] ?? []);
+
+        // Register middleware if messenger present
+        if (\class_exists(MessengerPass::class)) {
+            $middlewareDefinition = new Definition(ResetDecisionMiddleware::class);
+
+            $container->setDefinition(ResetDecisionMiddleware::class, $middlewareDefinition);
```
natepage, line 54: let's use a php config file to register the middleware so we can use autowire feature on it. Have a look at other packages to see how they use the `PhpFileLoader`.

packages/EasyDecision/src/Bridge/Symfony/Messenger/ResetDecisionMiddleware.php
```diff
@@ -0,0 +1,45 @@
...
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        if ($this->shouldSkip($envelope)) {
+            return $stack->next()
+                ->handle($envelope, $stack);
+        }
+
+        $this->decisionFactory->reset();
+
+        return $envelope;
```
natepage, line 34: Here you're bypassing all the other middleware, we can't do that 😄 in this case we MUST always return the envelope processed by the rest of middleware.

So the logic should be:

- if message consumed by worker, then reset decision factory
- return stack next handle

natepage (changes requested):

packages/EasyDecision/src/Bridge/Symfony/DependencyInjection/EasyDecisionExtension.php
```diff
@@ -43,5 +44,10 @@ public function load(array $configs, ContainerBuilder $container): void
         $container
             ->autowire(MappingProviderInterface::class, ConfigMappingProvider::class)
             ->setArgument('$decisionsConfig', $config['type_mapping'] ?? []);
+
+        // Register middleware if messenger present
+        if (\class_exists(MessengerPass::class)) {
+            $loader->load('messenger_client.php');
```
natepage, line 50: Can we please name this `messenger_middleware.php`, thank you, otherwise it looks good to me!

natepage (changes requested):

natepage (approved):

PR #428, opened 2020-11-28
[EasyBankFiles] Refactor DirectLink files parser.

ERuban:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | no    
| Fixed tickets | -   

Refactored DirectLink BankFiles parser a bit:

- Added errors array.
- Added one more check to getProcessedDate() method, to prevent throwing exception on `new DateTime()` with incorrect data.
- Make line code comparisons in the parser more strictly, to prevent a bug, e.g. `'w' == 0`.
- Added additional validation of the lines by its minimal valid length, otherwise one of our `substr()` can result a `false` value and we could have a TypeError exception in the application.

No tests implemented at this moment, seems it is messed up a bit in this package, - got a lot of errors:
```
Error: Class 'EoneoPay\Utils\Exceptions\ValidationException' not found
Error: Class 'EoneoPay\Utils\DateTime' not found
Error: Class 'Mockery' not found
```

I suggest to fix it in the separate PR with some additional refactoring here. (or I'm doing smth wrong with it ¯\_(ツ)_/¯ )

### DIFF
--- a/packages/EasyBankFiles/src/Parsers/DirectEntry/Parser.php
+++ b/packages/EasyBankFiles/src/Parsers/DirectEntry/Parser.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace EonX\EasyBankFiles\Parsers\DirectEntry;
 
 use EonX\EasyBankFiles\Parsers\AbstractLineByLineParser;
+use EonX\EasyBankFiles\Parsers\DirectEntry\Results\Error;
 use EonX\EasyBankFiles\Parsers\DirectEntry\Results\Header;
 use EonX\EasyBankFiles\Parsers\DirectEntry\Results\Trailer;
 use EonX\EasyBankFiles\Parsers\DirectEntry\Results\Transaction;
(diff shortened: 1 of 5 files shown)


roman-eonx (dismissed):

packages/EasyBankFiles/src/Parsers/DirectEntry/Parser.php
```diff
@@ -14,22 +15,42 @@ final class Parser extends AbstractLineByLineParser
     /**
      * @const Code for header line
      */
-    private const HEADER = 0;
+    private const HEADER = '0';
+
+    /**
+     * @const Minimal header line length (calculated from maximum substr arguments)
```
natepage, line 21: `@const` doesn't actually exist and it doesn't specify the actual type of value the constant holds. We MUST use `@var int`.

natepage (changes requested):

packages/EasyBankFiles/src/Parsers/DirectEntry/Parser.php
ERuban: Done.

natepage (approved):

Issue #431, opened 2020-12-05
[EasyErrorHandler] Incorrect translation for Laravel/Lumen setup


jayaregalinadaflexisourceit:
**Describe the bug**
All exceptions' default user message are unable to read because translations are not properly setup.

**To Reproduce**
Extend any exceptions from this package in Laravel/Lumen setup and try to throw any of that error into your application.

```php
// app\Exceptions\UnprocessableEntityException.php
final class UnprocessableEntityException extends ErrorException
{
    protected $statusCode = Response::HTTP_UNPROCESSABLE_ENTITY; // 422
}
```

Throw this exception in to your application

```php
...
throw new UnprocessableEntityException();
...
```

**Expected behavior**
All exceptions have their own default `$userMessage` property, depending on which exception. In the example above, a response should look like:

```
// STATUS: 422 Unprocessable Entity
{
  "code": 0,
  "message": "Oops, something went wrong.", // `exceptions.default_user_message`
  "sub_code": 0,
  "time": "2020-12-05T03:49:16Z"
}
```

**Additional context**
I think because of the `UserMessageBuilder` (from default builder) are only translating the user message without namespace. A couple of workarounds can be done here, like creating another UserMessageBuilder into your app or copying the `messages.php`  into your existing `resources/lang/{locale}/exceptions.php` translation.

natepage:
Done in #447

PR #432, opened 2020-12-05
[EasyErrorHandler] Fix/issue 431 incorrect translation for laravel or lumen setup

jayaregalinadaflexisourceit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | yes     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | #431


packages/EasyErrorHandler/src/Bridge/BridgeConstantsInterface.php
```diff
@@ -6,6 +6,11 @@
 
 interface BridgeConstantsInterface
 {
+    /**
+     * @var string
+     */
+    public const TRANSLATION_NAMESPACE = 'easy-error-handler';
```
natepage, line 12: Const MUST be alpha ordered so this one should be at the very bottom of the file.

packages/EasyErrorHandler/src/Bridge/Laravel/Provider/EasyErrorHandlerServiceProvider.php
```diff
@@ -35,20 +36,27 @@ public function register(): void
 
         $this->app->singleton(ErrorResponseFactoryInterface::class, ErrorResponseFactory::class);
 
+        $this->app->singleton(ErrorHandlerInterface::class, function (): ErrorHandlerInterface {
```
natepage, line 39: We should recreate a new PR so we don't have the updates from the other PR here so we avoid any conflicts, especially because we don't need those changes to update the translation logic.

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
```diff
@@ -14,19 +15,37 @@ final class Translator implements TranslatorInterface
...
+    private $translationNamespace;
+
+    public function __construct(IlluminateTranslatorInterface $decorated, string $translationNamespace)
     {
         $this->decorated = $decorated;
+        $this->translationNamespace = $translationNamespace;
     }
 
     /**
+     * @param string $message
```
natepage, line 30: No need for these because we have type hint already. PHPDoc are required only for arguments/return that are not explicit about their content (e.g. array, iterable, ...)

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
```diff
@@ -14,19 +15,37 @@ final class Translator implements TranslatorInterface
...
+     *
+     * @return string
      */
     public function trans(string $message, array $parameters): string
     {
         // TODO: rework after upgrading all the illuminate and laravel packages to ^6.0
         $method = \method_exists($this->decorated, 'trans') ? 'trans' : 'get';
 
+        $translation = $this->decorated->{$method}(
+            \implode([$this->translationNamespace, '::', \trim($message)]),
```
natepage, line 41: Use `\sprintf()` instead of `\implode()` array are slow and really complex for this use case.

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
```diff
@@ -14,19 +15,37 @@ final class Translator implements TranslatorInterface
...
+            $parameters
+        );
+
+        if ($translation !== $message && Str::startsWith($this->translationNamespace . '::', $translation)) {
```
natepage, line 45: Checking if the translated message starts with the translation namespace for early return doesn't make sense, because if it actually translated the message properly then it won't have the namespace in the result.

Also we should try to translate without the namespace first, and default to the namespace because we should have more messages without the namespace than messages with namespace.

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Laravel/Provider/EasyErrorHandlerServiceProvider.php
jayaregalinadaflexisourceit: I created a separate PR for this #438

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
jayaregalinadaflexisourceit: Thanks, I'll use `sprintf` instead

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
jayaregalinadaflexisourceit: Thank you, I'll note of this

jayaregalinadaflexisourceit:
I'll close this now and recreate a new PR soon related to the translation to avoid future conflicts

PR #433, opened 2020-12-07
[EasyErrorHandler] Allow ignore exceptions in bugnsag reporter provider

jordancachero:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes 
| Fixed tickets |


packages/EasyErrorHandler/src/Bridge/Laravel/config/easy-error-handler.php
```diff
@@ -23,6 +25,10 @@
      * Log level threshold to notify bugsnag.
      */
     'bugsnag_threshold' => \env('EASY_ERROR_HANDLER_EASY_BUGSNAG_ENABLED', null),
+    /**
+     * List of Ignored Exceptions that'll not be reported to bugsnag.
+     */
+    'bugsnag_ignored_exceptions' => [HttpExceptionInterface::class],
```
natepage, line 31: The default value is already handled in the `BugsnagReporter` so the default value of the config should be null.

natepage (changes requested):

natepage:
Merged already in another PR, thanks.

PR #447, opened 2020-12-15
[EasyErrorHandler] Fix translation setup in Laravel/Lumen setup

jayaregalinadaflexisourceit:
- Renamed messages to exceptions translation file
- Translate with namespace
- Added `TRANSLATION_NAMESPACE` constant

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | #431 #432


packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
```diff
@@ -24,8 +25,24 @@ public function __construct(IlluminateTranslatorInterface $decorated)
...
+        return $this->getOrLang(
+            \sprintf('%s::%s', BridgeConstantsInterface::TRANSLATION_NAMESPACE, $message),
+            $parameters
+        );
+    }
+
+    /**
+     * @param mixed[] $parameters
+     */
+    private function getOrLang(string $message, array $parameters): string
```
natepage, line 43: Let's call this method `doTrans()` so it doesn't sound like a getter 😄

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
```diff
@@ -24,8 +25,24 @@ public function __construct(IlluminateTranslatorInterface $decorated)
...
     {
-        // TODO: rework after upgrading all the illuminate and laravel packages to ^6.0
-        $method = \method_exists($this->decorated, 'trans') ? 'trans' : 'get';
+        $translation = $this->getOrLang($message, $parameters);
+
+        if ($translation !== $message) {
+            return $translation;
+        }
+
+        return $this->getOrLang(
```
natepage, line 34: What happen if the translation doesn't exist even in the package namespace?

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
jayaregalinadaflexisourceit: Like as usual, it will return as it is, however with a namespace on it. Like:
```json
{
  "message": "easy-error-handler:non-existing-error",
  "subcode": 0,
  ...
}
```

natepage: @jayaregalinadaflexisourceit ok do you think that's a good solution? 😄

jayaregalinadaflexisourceit: hmmm, according to the test, I think nope, I think I should the `startWith` checker here.

natepage: @jayaregalinadaflexisourceit ok I agree it's not a good solution haha

But no need for `startWith`, just assign the message prefixed with the namespace to a variable, proceed with translation and if the result is the same as the variable you created then you return the original message 😉

jayaregalinadaflexisourceit: Ok, I think that is even better, and less dependent. I'll try that now

packages/EasyErrorHandler/src/Bridge/Laravel/Translator.php
```diff
@@ -24,9 +25,37 @@ public function __construct(IlluminateTranslatorInterface $decorated)
      */
     public function trans(string $message, array $parameters): string
     {
-        // TODO: rework after upgrading all the illuminate and laravel packages to ^6.0
-        $method = \method_exists($this->decorated, 'trans') ? 'trans' : 'get';
+        $translation = $this->doTrans($message, $parameters, false);
```
natepage, line 28: Here is what I would do to keep things simple to understand/read.

- Delete the `processMessage()` method we don't need it
- Revert the changes made to `doTrans()` this method should have only 1 purpose, translate

And then that's how things would look like:

```php
$translation = $this->doTrans($message, $parameters);

if ($translation !== $message) {
	return $translation;
}

$namespacedMessage = \sprintf('%s::%s', BridgeConstantsInterface::TRANSLATION_NAMESPACE, $message);
$translation = $this->doTrans($namespacedMessage, $parameters);

// If translation is finally different we return it otherwise default to original message
return $translation !== $namespacedMessage ? $translation : $message; 
```

natepage (changes requested):

PR #461, opened 2021-01-13
[EasyErrorHandler] Introduced ApiPlatformBuilderProvider and ApiPlatformValidationExceptionBuilder

itorgov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


itorgov:
Actually it's not ready yet.

itorgov:
Now it's ready :-)

packages/EasyErrorHandler/src/Bridge/Symfony/Builders/ApiPlatformBuilderProvider.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Builders;
```
roman-eonx, line 5: We use singular namespaces in Symfony, so this should be `EonX\EasyErrorHandler\Bridge\Symfony\Builder`.

roman-eonx (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Builders/ApiPlatformBuilderProvider.php
itorgov: Fixed.

roman-eonx (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public function buildData(Throwable $throwable, array $data): array
+    {
+        if ($throwable instanceof ValidationException) {
+            $violations = [];
+
+            foreach ($throwable->getConstraintViolationList() as $violation) {
+                $propertyPath = $violation->getPropertyPath();
+
+                if (\array_key_exists($propertyPath, $violations) === false) {
```
natepage, line 48: Using `isset()` here would be better.

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+                }
+
+                $violations[$propertyPath][] = $violation->getMessage();
+            }
+
+            $exceptionKey = $this->getKey('exception');
+            $exceptionMessageKey = $this->getKey('extended_exception_keys.message');
+
+            if (\is_array($data[$exceptionKey] ?? null)) {
+                $data[$exceptionKey][$exceptionMessageKey] = 'Entity validation failed.';
```
natepage, line 59: Not every project might want to have this generic message `Entity validation failed.`, can we please make it configurable.

packages/EasyErrorHandler/src/Bridge/Laravel/translations/en/exceptions.php
```diff
@@ -8,5 +8,6 @@
     'conflict' => 'Conflict.',
     'forbidden' => 'Forbidden.',
     'not_found' => 'Not found.',
+    'not_valid' => 'Validation failed.',
```
natepage, line 11: This is the Laravel translations file 😄

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
natepage: Actually it could just be a translation.

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
```diff
@@ -0,0 +1,98 @@
...
+        $keys = $keys ?? $this->keys;
+        $nameParts = \explode('.', $name);
+
+        if (\count($nameParts) <= 1) {
+            return $keys[$name] ?? $name;
+        }
+
+        $firstPartOfName = \array_shift($nameParts);
+
+        if (\array_key_exists($firstPartOfName, $keys) === false) {
```
natepage, line 92: Please use `isset()` here as well.

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/api_platform_builders.php
```diff
@@ -0,0 +1,23 @@
...
+        ->autoconfigure();
+
+    $services
+        ->set(ApiPlatformBuilderProvider::class)
+        ->arg('$keys', '%' . BridgeConstantsInterface::PARAM_RESPONSE_KEYS . '%');
+
+    // Drop tags of ValidationExceptionListener
+    $services
+        ->set('api_platform.listener.exception.validation')
+        ->synthetic(true);
```
natepage, line 22: I'm not sure about that, I had issues in the past because a `synthetic` service must be initialised so if something else needs it, it might break the application...

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
itorgov: I'm not sure that it has to be a translatable. It's a message for the exception and it won't be visible for a user on production. Anyway, do you have any suggestions for the translation key?

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/api_platform_builders.php
itorgov: I didn't find any other ways to remove tags from this service. There was an another option  - define it as abstract (`->abstract(true)`) but I didn't tested this solution.

packages/EasyErrorHandler/src/Bridge/Laravel/translations/en/exceptions.php
itorgov: I've updated the default user message for the `packages/EasyErrorHandler/src/Exceptions/ValidationException.php` exception that's why I added it here :-)

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/api_platform_builders.php
itorgov: I've made it abstract and what I've gotten:
image

As you can see tags and a class was removed.

itorgov: Btw, it's really close to how it was with synthetic:
image

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
itorgov: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
itorgov: Done.

roman-eonx (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/api_platform_builders.php
natepage: @itorgov did you try this solution in `payment-gateway-php` project to see if it works?

Another cleaner solution would be to create a compiler pass that completely remove this definition from the container, more verbose but I am sure it works.

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
natepage: @itorgov we already have the translation key: `extended_exception_keys.message`

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php
itorgov: Done by using another key :-)

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/api_platform_builders.php
itorgov: Finally I managed to remove that listener by using a compiler pass. Thanks for the suggestion!

roman-eonx (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/Configuration.php
```diff
@@ -15,6 +15,7 @@ public function getConfigTreeBuilder(): TreeBuilder
 
         $treeBuilder->getRootNode()
             ->children()
+                ->booleanNode('api_platform')->defaultTrue()->end()
```
natepage, line 18: @itorgov ok so if the intention of this config/param is to allow devs to keep using the default api platform listener, let's make it more explicit and call it something like `override_api_platform_listener` 😉

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/EasyErrorHandlerExtension.php
```diff
@@ -25,6 +25,8 @@ public function load(array $configs, ContainerBuilder $container): void
         $config = $this->processConfiguration(new Configuration(), $configs);
         $loader = new PhpFileLoader($container, new FileLocator([__DIR__ . '/../Resources/config']));
 
+        $container->setParameter(BridgeConstantsInterface::PARAM_IS_API_PLATFORM, $config['api_platform']);
```
natepage, line 28: @itorgov here same, let's make sure we update the constant/parameter accordingly to the new config name.

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/Configuration.php
itorgov: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/EasyErrorHandlerExtension.php
itorgov: Done.

roman-eonx (approved):

natepage (approved):

PR #462, opened 2021-01-18
[EasyStandard] Implement AvoidPrivatePropertiesSniff

DKeeper:
Added `AvoidPrivatePropertiesSniff`, a copy of `AvoidPublicPropertiesSniff`. This new sniff will help us to keep only protected properties in entity classes (as we decided to not use private properties there because of inheritance issues).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |

### DIFF
--- a/packages/EasyStandard/docs/ecs/sniffs.md
+++ b/packages/EasyStandard/docs/ecs/sniffs.md
@@ -5,7 +5,7 @@ weight: 1001
 #### Classes
 
 ##### [\EonX\EasyStandard\Sniffs\Classes\AvoidPrivatePropertiesSniff][1]
-Class properties must be protected or private.
+Class properties must be protected or public.
 ```php
 // incorrect
 class MyClass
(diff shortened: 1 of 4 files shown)


packages/EasyStandard/docs/ecs/sniffs.md
```diff
@@ -4,7 +4,28 @@ weight: 1001
 ---eonx_docs---
 #### Classes
 
-##### [\EonX\EasyStandard\Sniffs\Classes\AvoidPublicPropertiesSniff][1]
+##### [\EonX\EasyStandard\Sniffs\Classes\AvoidPrivatePropertiesSniff][1]
+Class properties must be protected or private.
```
roman-eonx, line 8: It should be `or public` here, right?

roman-eonx (changes requested):

PR #488, opened 2021-02-26
[EasyWebhook] Improve docs

nick-lc:
Draft EasyWebhook docs. I tried to cover everything I could think of... Please let me know what I missed / overdid / got wrong!

### DIFF
--- a/packages/EasyWebhook/docs/middleware.md
+++ b/packages/EasyWebhook/docs/middleware.md
@@ -9,19 +9,19 @@ The EasyWebhook package uses a **middleware stack** that is processed every time
 middleware proceeds down the stack to where the webhook is sent in a HTTP request, and then proceeds back up the stack.
 Actions may be taken on the way down and/or the way back up.
 
-Middleware are services implementing `EonX\EasyWebhook\Interfaces\MiddlewareInterface`. You can write custom middleware,
-e.g. to add a custom header to webhook HTTP requests.
+Middleware are services implementing `EonX\EasyWebhook\Interfaces\MiddlewareInterface`. This package provides **core
+middleware**, but you can also write **custom middleware**, e.g. to add a custom header to webhook HTTP requests.
 
 The middleware is ordered by **priority** in the stack, whereby the lower the priority, the earlier in the stack the
 middleware is placed.
 
-The core middleware can have priorities between -5000 and 5000. If not specified, the default priority is 0. Some
-middleware provided by this package must run before the core middleware, e.g. to lock the Webhook object, and some
-middleware must run after the core middleware, e.g. to send the webhook HTTP request.
+Custom middleware can have priorities between -5000 and 5000. If not specified, the default priority is 0. Some core
+middleware provided by this package must run before the custom middleware, e.g. to lock the Webhook object, and some
+core middleware must run after the core middleware, e.g. to send the webhook HTTP request.
 
 ## Configure once middleware
 
-Some middleware is **configure once** middleware. If the webhook is flagged as **configured**, i.e. its `$configured`
+Some middleware are **configure once** middleware. If the webhook is flagged as **configured**, i.e. its `$configured`
 property is `true`, then configure once middleware will not operate on the webhook.
 
 Note that if a webhook is retrieved from a webhook store, then it is automatically flagged as configured.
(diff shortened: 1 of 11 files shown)


packages/EasyWebhook/docs/middleware.md
```diff
@@ -0,0 +1,217 @@
...
+able to find due webhooks. Of the stores provided by the EasyWebhook package, only the Doctrine DBAL webhook store
+supports finding due webhooks.
+
+### `SendWebhookMiddleware`
+
+This middleware sends the webhook HTTP request and returns the result in a [WebhookResult](webhook-result.md) object.
+
+### `SignatureHeaderMiddleware`
+
+This middleware sets the `X-Webhook-Signature` header of the webhook HTTP request. The header contains a SHA-256 HMAC
```
natepage, line 108: Here we should say that the signature logic can be customised by creating a new implementation of the signature service and telling the package to use it.
And also that the header name is configurable.

packages/EasyWebhook/docs/middleware.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Middleware stack
+
+The following table show the middleware stack in priority order, with summaries of their actions:
+
+| Middleware | Action forward :arrow_down: | Action back :arrow_up: |
+| ---------- | --------------------------- | ---------------------- |
+| `LockMiddleware` | Lock webhook | Unlock webhook |
+| `RerunMiddleware` | If rerun allowed, reset status and current attempt<br/>If not allowed, throw exception | |
+| *Begin core middleware*<br/>*(processed in priority order)* | | |
```
natepage, line 141: Like I said in a comment above, "core middleware" are the one from the package that CANNOT be removed to guarantee the package to work as expected.

So here, the core middleware don't begin but end 😛

natepage: Here the "Before Core Middleware" end.

packages/EasyWebhook/docs/middleware.md
```diff
@@ -0,0 +1,217 @@
...
+| `BodyFormatterMiddleware`| Format request body | |
+| `EventHeaderMiddleware` | Set `X-Webhook-Event` request header | |
+| `IdHeaderMiddleware` | Set `X-Webhook-Id` request header | |
+| `SignatureHeaderMiddleware` | Set `X-Webhook-Signature` request header | |
+| Custom middleware | Custom pre-processing | Custom post-processing |
+| *End core middleware* | | |
```
natepage, line 147: Here is the "After Core Middleware" start.

packages/EasyWebhook/docs/middleware.md
```diff
@@ -0,0 +1,217 @@
...
+| `MethodMiddleware` | Set request method | |
+| `SendAfterMiddleware` | If time is before `$sendAfter`, store webhook and return up stack<br/>If time is after `$sendAfter`, continue down stack | |
+| `AsyncMiddleware` | If asynchronous, store webhook and return up stack<br/>If synchronous, continue down stack | |
+| `StoreMiddleware` | | Store webhook and result |
+| `EventsMiddleware` | | Dispatch event |
+| `StatusAndAttemptMiddleware` | | Update status and attempt |
+| `SendWebhookMiddleware` | Send request and return up stack | |
+
+::: warning FIXME
+Priority of `SendAfterMiddleware`?
```
natepage, line 158: Yeah, I've checked the code and we're missing the `SendAfterMiddleware` 😄 will have to fix that.

packages/EasyWebhook/docs/webhook.md
```diff
@@ -0,0 +1,129 @@
...
+# Webhook class
+
+The **Webhook** class defines webhook configuration. Once configured, the [WebhookClient](webhook-client.md) class is
+responsible for sending the webhook.
+
+## Properties
+
+A Webhook object has the following properties:
+
+- `$allowRerun`: Boolean indicating whether a webhook can be rerun multiple times.
```
natepage, line 15: To not confuse anyone with "rerun" and "retry", we should specify here that "can be rerun multiple times after being successful/failed".

packages/EasyWebhook/docs/webhook.md
```diff
@@ -0,0 +1,129 @@
...
+- `$body`: Body of the webhook HTTP request, provided as an array of key-value pairs, which will be formatted as a JSON
```
natepage, line 16: It won't be necessarily formatted to JSON, it depends on the body formatter service used.

packages/EasyWebhook/docs/webhook.md
```diff
@@ -0,0 +1,129 @@
...
+  string in the request.
+- `$bodyAsString`: Body of the webhook HTTP request, provided as a JSON string, e.g. `{"event":"showcase"}`. Note that
```
natepage, line 18: It doesn't have to be JSON either, if people want to set XML in here they can.

natepage (changes requested):

natepage (approved):

Issue #510, opened 2021-03-25
[EasyWebhook] Sync retry


natepage:
Failing webhooks are only retried when sent asynchronously, and for good reasons.
However it might be handy to implement a fallback retry strategy for sync sending as well.

natepage:
Done in #517 .

PR #517, opened 2021-03-29
[EasyWebhook] Implement synchronous retries

natepage:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | #510

### DIFF
--- a/packages/EasyWebhook/docs/middleware.md
+++ b/packages/EasyWebhook/docs/middleware.md
@@ -158,6 +158,11 @@ The status can be set to one of:
 This middleware stores the webhook and webhook result in the configured stores after a response has been received from
 the webhook HTTP request. See [Stores](stores.md) for more information.
 
+### `SyncRetryMiddleware`
+
+This middleware retries sending the webhook when sent synchronously, it provides a simple solution out-the-box to handle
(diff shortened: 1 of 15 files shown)


packages/EasyWebhook/docs/middleware.md
```diff
@@ -158,6 +158,11 @@ The status can be set to one of:
 This middleware stores the webhook and webhook result in the configured stores after a response has been received from
 the webhook HTTP request. See [Stores](stores.md) for more information.
 
+### `SyncRetryMiddleware`
+
+This middleware retries sending the webhook when sent synchronously, it provides a simple solution out-the-box to handle
+retries. However, we strongly recommend using the async functionalities to use proper retry strategy.
```
nick-lc, line 164: I would reword this a bit. Maybe something like:

> If the webhook was sent synchronously and it failed, this middleware retries sending the webhook. It provides a simple out-of-the-box solution for handling retries. However, we strongly recommend sending webhooks asynchronously so your application is not blocked by retries.

nick-lc (changes requested):
Suggested some rewording :)

nick-lc (approved):
I'm happy :)

PR #567, opened 2021-05-12
[easy-monorepo] Remove EasyStandard and add EasyQuality for quality checks

ERuban:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes 
| Deprecations? | no
| Tests pass?   | yes 
| Fixed tickets | -

- Remove EasyStandard
- Remove `quality` dependencies
- Add EasyQuality

We have some `ECS` issues, but I think it will be fixed after updating `ECS` in `easy-quality` package.


packages/EasyApiToken/composer.json
```diff
@@ -16,7 +16,7 @@
         "auth0/auth0-php": "^5.4",
         "psr/container": "^1.0",
         "laravel/lumen-framework": "^5.5",
-        "symfony/dependency-injection": "5.1.10",
+        "symfony/dependency-injection": "^5.1.10",
```
roman-eonx, line 19: It was `^4.4 || ^5.1.5` [before the 3.0.5 version](https://github.com/eonx-com/easy-monorepo/blob/3.0.4/packages/EasyApiToken/composer.json). Let's revert it to that restriction back.

.github/workflows/php-ci.yml
```diff
@@ -36,6 +36,8 @@ jobs:
 
             -   uses: eonx-com/actions-composer-install@v1
 
+            -   run: composer install
```
roman-eonx, line 39: If this is needed only for installing EasyQuality, then let's add `--working-dir=.quality`.

rector-ci.php
```diff
@@ -9,7 +9,6 @@
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
-    $containerConfigurator->import(__DIR__ . '/packages/EasyStandard/config/rector-set.php');
```
roman-eonx, line 12: This way we stop using our set in CI. I think we should introduce a new set for this (right in the monorepo) or simply move the contents of that set here.

roman-eonx (changes requested):

packages/EasyApiToken/composer.json
ERuban: done

.github/workflows/php-ci.yml
ERuban: done

rector-ci.php
ERuban: added to config, but rector checks are disabled in php-ci

roman-eonx (approved):

natepage (approved):
Thank you for that!

Issue #585, opened 2021-06-08
[EasyErrorHanlder] Add DataCollector (symfony)


ERuban:
Add DataCollector to the `EasyErrorHandler` package to make possible asserting reported data using profiler.

BOB41K1987:
@ERuban @natepage As far as I know, the profiler wouldn't work in console commands and messenger handlers. I'd suggest adding in test environment `InMemoryReporter implements ErrorReporterInterface` the sole purpose of which would be to collect reported exceptions and lets us assert them afterward.

natepage:
@BOB41K1987 it would work but you're right, if the only use case is to assert things in tests, then simply decorate the ErrorHandlerInterface in test environment and add getters to it so you can fetch the errors at a later stage.

ERuban:
@natepage yes, the only use case at this moment is possibility to assert reported exceptions.

ERuban:
So, I've tried to decorate it, and it works in places I want, but seems that ErrorHandler::render() became broken - other tests with the error responses now gets an empty array as a response.
At this moment it looks like:
```
    # services_test.yaml
    EonX\EasyErrorHandler\ErrorHandler:
        arguments:
            $builderProviders: !tagged_iterator \EonX\EasyErrorHandler\Bridge\BridgeConstantsInterface::TAG_ERROR_RESPONSE_BUILDER_PROVIDER
            $reporterProviders: !tagged_iterator \EonX\EasyErrorHandler\Bridge\BridgeConstantsInterface::TAG_ERROR_REPORTER_PROVIDER

    EonX\EasyErrorHandler\Interfaces\ErrorHandlerInterface:
        class: App\Tests\Stub\ErrorHandlerStub
        decorates: EonX\EasyErrorHandler\ErrorHandler
```
(Somehow it doesn't want to work without explicity configured value - `argument "$builderProviders" of method "EonX\EasyErrorHandler\ErrorHandler::__construct()" is type-hinted "iterable", you should configure its value explicitly`)

And the stub-decorator code:
```
final class ErrorHandlerStub implements ErrorHandlerInterface
{
    private ErrorHandlerInterface $decorated;

    private ?Throwable $throwable = null;

    public function __construct(ErrorHandlerInterface $decorated)
    {
        $this->decorated = $decorated;
    }

    public function getReportedException(): ?Throwable
    {
        return $this->throwable;
    }

    public function isVerbose(): bool
    {
        return $this->decorated->isVerbose();
    }

[...]

PR #587, opened 2021-06-10
[EasyErrorHandler] Add Symfony Profiler DataCollector

ERuban:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #585

Add Symfony's Profiler DataCollector to collect reported exceptions to simplify functional tests.


packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/TraceableErrorHandlerInterface.php
```diff
@@ -0,0 +1,13 @@
...
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Interfaces;
+
+use EonX\EasyErrorHandler\Interfaces\ErrorHandlerInterface;
+use Throwable;
+
+interface TraceableErrorHandlerInterface extends ErrorHandlerInterface
+{
+    public function getReportedException(): ?Throwable;
```
natepage, line 12: @ERuban applications could report more than one error per request, so this should return an array.

Also the class is called "ErrorHandlerInterface" so this method should say "getReportedErrors()".

And last one, what about the rendered errors? Applications could use the render method to render more than one errors without necessarily returning them to the consumers, and we might want to collect them too.

natepage (changes requested):

natepage:
@ERuban also, if we really go down the path of implementing a data collector, then I think we should do properly (template, etc.) might be useful.

That said, it would be great to collect the list of ResponseDataBuilders and Reporters so we can show them in the profiler with their class and priority :wink:

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/TraceableErrorHandlerInterface.php
ERuban: Done

roman-eonx (approved):

packages/EasyErrorHandler/src/Interfaces/ErrorHandlerInterface.php
```diff
@@ -10,6 +10,16 @@
 
 interface ErrorHandlerInterface
 {
+    /**
+     * @return \EonX\EasyErrorHandler\Interfaces\ErrorResponseBuilderInterface[]
+     */
+    public function getBuilders(): array;
```
natepage, line 16: I don't think we should require those methods to be implemented by all ErrorHanlders when it's only a thing we need for symfony in dev/test environment.

packages/EasyErrorHandler/src/ErrorHandler.php
```diff
@@ -55,6 +55,16 @@ public function __construct(
         $this->isVerbose = $isVerbose ?? false;
     }
 
+    public function getBuilders(): array
```
natepage, line 58: PHPDoc is missing.

packages/EasyErrorHandler/src/ErrorHandler.php
```diff
@@ -55,6 +55,16 @@ public function __construct(
...
+    {
+        return $this->builders;
+    }
+
+    public function getReporters(): array
```
natepage, line 63: PHPDoc is missing.

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/Compiler/ApiPlatformCompilerPass.php
```diff
@@ -5,13 +5,24 @@
...
+use EonX\EasyErrorHandler\Interfaces\ErrorHandlerInterface;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
 
 class ApiPlatformCompilerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
+        if ($container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug')) {
```
natepage, line 19: This has nothing to do with ApiPlatform, it shouldn't be in this compiler pass.

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/Compiler/ApiPlatformCompilerPass.php
ERuban: `EasyErrorHandlerSymfonyCompilerPass` naming will be ok here?

packages/EasyErrorHandler/src/Interfaces/ErrorHandlerInterface.php
ERuban: Yeap, agree with you. Done it with `@method` annotation in the `ErrorHandlerInterface` and moving it to the `TraceableErrorHandlerInterface`. What do you think?

packages/EasyErrorHandler/src/ErrorHandler.php
ERuban: Done

packages/EasyErrorHandler/src/ErrorHandler.php
ERuban: Done

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/Compiler/ApiPlatformCompilerPass.php
ERuban: Done it with `ErrorHandlerCompilerPass` name

PR #591, opened 2021-06-19
[EasyErrorHandler] Add an alias to ErrorHandler to allow service decoration

BOB41K1987:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


natepage:
Sorry, maybe I'm missing something but we shouldn't have to create an alias to be able to decorate a service. I think we should take some time to discuss that together over a call and get to the bottom of it.

natepage:
@BOB41K1987 thank you for the PR, as discussed, this is not needed as we can decorate the service without this alias.

PR #676, opened 2021-07-29
[EasyBatch] Fix Processed increment on BatchItem cancelling

ERuban:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets |    

- Fix Processed increment on BatchItem cancelling.


natepage:
@ERuban thank you for your PR, we definitely need to solve this issue.

However, your current implementation isn't gonna work, each computed number on the batch must be updated atomically, so we need to tweak your solution a bit to use the `updateBatchForItem()` instead.

natepage (approved):

PR #733, opened 2021-09-14
[EasyCore] Add new constraints

alexndlm:
Add new constraints:
- Alphanumeric
- AlphanumericHyphen
- DateInterval
- Decimal

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphen.php
+++ b/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphen.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyCore\Bridge\Symfony\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ * @Target({"PROPERTY", "METHOD", "ANNOTATION"})
+ */
+final class AlphanumericHyphen extends Constraint
+{
+    /**
+     * @var string
+     */
+    public const INVALID_ALPHANUMERIC_HYPHEN_ERROR = 'INVALID_ALPHANUMERIC_HYPHEN_ERROR';
+
+    /**
+     * @var string
+     */
+    public $message = 'This value may only contain letters, numbers and hyphens.';
+}
--- a/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphen.php
+++ b/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphen.php
@@ -20,5 +20,5 @@ final class AlphanumericHyphen extends Constraint
     /**
      * @var string
      */
-    public $message = 'This value may only contain letters, numbers and hyphens.';
+    public $message = 'This value may only contain letters, numbers, and hyphens.';
 }
--- a/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphenValidator.php
+++ b/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphenValidator.php
@@ -13,7 +13,7 @@ final class AlphanumericHyphenValidator extends ConstraintValidator
 {
     public function validate($value, Constraint $constraint)
     {
-        if (!$constraint instanceof AlphanumericHyphen) {
+        if ($constraint instanceof AlphanumericHyphen === false) {
             throw new UnexpectedTypeException($constraint, AlphanumericHyphen::class);
         }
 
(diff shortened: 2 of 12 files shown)


packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphen.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * @var string
+     */
+    public const INVALID_ALPHANUMERIC_HYPHEN_ERROR = 'INVALID_ALPHANUMERIC_HYPHEN_ERROR';
+
+    /**
+     * @var string
+     */
+    public $message = 'This value may only contain letters, numbers and hyphens.';
```
roman-eonx, line 23:
```suggestion
    public $message = 'This value may only contain letters, numbers, and hyphens.';
```

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/Decimal.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * @var int
+     */
+    public $maxPrecision;
+
+    /**
+     * @var string
+     */
+    public $message = 'This value is not a valid decimal number or has ' .
+    'not {{ minPrecision }} - {{ maxPrecision }} digits in precision.';
```
roman-eonx, line 30:
```suggestion
    public $message = 'This value is not a valid decimal or integer number, has ' .
    'less than {{ minPrecision }} or more than {{ maxPrecision }} digits in precision.';
```

roman-eonx (changes requested):

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphen.php
alexndlm: Done.

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/Decimal.php
alexndlm: Done.

roman-eonx (approved):

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphenValidator.php
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        if (\is_scalar($value) === false
+            && (\is_object($value) && \method_exists($value, '__toString')) === false) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        $value = (string)$value;
+
+        if (\preg_match('/^[a-z0-9_-]+$/i', $value) === 0) {
```
voodooism, line 31: The message in the constraint class says that only alphanumeric and hyphens are allowed, but I see an underscore symbol in the regexp here. Is it okay?

voodooism (dismissed):

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/AlphanumericHyphenValidator.php
alexndlm: I think that in this case, it is acceptable.
@roman-eonx - what do you think?

voodooism: I have nothing against it, but in this case, I think it would be a good idea to extend the message in the constraint class

roman-eonx: Yes, hyphens = underscores + minuses in this case (as our technical writer decided). And the validation message was updated by him from the old "dashes" one.

packages/EasyCore/tests/Bridge/Symfony/Validator/Constraints/DateIntervalValidatorTest.php
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * @return mixed[]
+     *
+     * @see testValidateThrowsUnexpectedValueException
+     */
+    public function provideUnexpectedValues(): array
+    {
+        return [
+            'integer' => [53004085616, 'Expected argument of type "string", "integer" given'],
```
itorgov, line 45: I think we don't need a data provider for only one dataset.

itorgov (changes requested):

packages/EasyCore/tests/Bridge/Symfony/Validator/Constraints/DateIntervalValidatorTest.php
alexndlm: Agree. Deleted.

roman-eonx (approved):

itorgov (approved):

voodooism (approved):

natepage (approved):

PR #734, opened 2021-09-15
[EasyCore] Make sure DeferredEntityEventDispatcher dispatches one event per entity

mvlad:
In the easy-core package we have the `\EonX\EasyCore\Doctrine\Dispatchers\DeferredEntityEventDispatcher` event dispatcher that allows us to dispatch the `\EonX\EasyCore\Doctrine\Events\EntityCreatedEvent` and `\EonX\EasyCore\Doctrine\Events\EntityUpdatedEvent` events when an entity really saved in a database. But we have found a problem with this dispatcher.

Let’s imagine the following flow:
* Begin a transaction.
* Create an entity and flush it.
* Update the entity and flush.
* Update the entity again and flush.
* Commit the transaction.

When the transaction will be committed our dispatcher will dispatch 3 events: one `EntityCreatedEvent` and two `EntityUpdatedEvent`. It’s wrong because in terms of the database we have created the entity and that's all. The same for the regular updates. If we update and flush some entity many times inside a transaction then we should dispatch only one `EntityUpdatedEvent` event after the transaction will be committed.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
+++ b/packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
@@ -102,13 +102,23 @@ final class DeferredEntityEventDispatcher implements DeferredEntityEventDispatch
             return;
         }
 
-        \array_walk_recursive($entityInsertions, function (object $entity): void {
-            $this->eventDispatcher->dispatch(new EntityCreatedEvent($entity));
-        });
+        $processedEntities = [];
+        foreach ($entityInsertions as $entities) {
+            foreach ($entities as $oid => $entity) {
+                $processedEntities[$oid] = $entity;
+                $this->eventDispatcher->dispatch(new EntityCreatedEvent($entity));
+            }
+        }
 
-        \array_walk_recursive($entityUpdates, function (object $entity): void {
-            $this->eventDispatcher->dispatch(new EntityUpdatedEvent($entity));
-        });
+        foreach ($entityUpdates as $entities) {
+            foreach ($entities as $oid => $entity) {
+                if (\array_key_exists($oid, $processedEntities)) {
+                    continue;
+                }
+                $processedEntities[$oid] = $entity;
+                $this->eventDispatcher->dispatch(new EntityUpdatedEvent($entity));
+            }
+        }
     }
 
     public function enable(): void
--- a/packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
+++ b/packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
@@ -112,7 +112,7 @@ final class DeferredEntityEventDispatcher implements DeferredEntityEventDispatch
 
         foreach ($entityUpdates as $entities) {
             foreach ($entities as $oid => $entity) {
-                if (\array_key_exists($oid, $processedEntities)) {
+                if (isset($processedEntities[$oid])) {
                     continue;
                 }
                 $processedEntities[$oid] = $entity;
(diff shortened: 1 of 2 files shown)


packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -102,13 +102,23 @@ public function dispatch(): void
...
+                $this->eventDispatcher->dispatch(new EntityCreatedEvent($entity));
+            }
+        }
 
-        \array_walk_recursive($entityUpdates, function (object $entity): void {
-            $this->eventDispatcher->dispatch(new EntityUpdatedEvent($entity));
-        });
+        foreach ($entityUpdates as $entities) {
+            foreach ($entities as $oid => $entity) {
+                if (\array_key_exists($oid, $processedEntities)) {
```
itorgov, line 115: Please, use `isset` instead.

packages/EasyCore/tests/Doctrine/Dispatchers/DeferredEntityEventDispatcherTest.php
```diff
@@ -266,6 +268,59 @@ public function testDispatchSucceedsWithEntities(): void
...
+    {
+        /** @var object $newEntity */
+        $newEntity = $this->prophesize(stdClass::class)->reveal();
+        $oid = \spl_object_hash($newEntity);
+        $entityCreatedEvent = new EntityCreatedEvent($newEntity);
+        $entityUpdatedEvent = new EntityUpdatedEvent($newEntity);
+        $eventDispatcher = $this->prophesize(EventDispatcherInterface::class);
+        $eventDispatcher->dispatch($entityCreatedEvent)
+            ->willReturnArgument(0);
+        $eventDispatcher->dispatch($entityUpdatedEvent)
```
itorgov, line 281: I'm not sure that we even need to define this.

packages/EasyCore/tests/Doctrine/Dispatchers/DeferredEntityEventDispatcherTest.php
```diff
@@ -266,6 +268,59 @@ public function testDispatchSucceedsWithEntities(): void
...
+    {
+        /** @var object $existedEntity */
+        $existedEntity = $this->prophesize(stdClass::class)->reveal();
+        $oid = \spl_object_hash($existedEntity);
+        $entityUpdatedEvent = new EntityUpdatedEvent($existedEntity);
+        $eventDispatcher = $this->prophesize(EventDispatcherInterface::class);
+        $eventDispatcher->dispatch($entityUpdatedEvent)
+            ->willReturnArgument(0);
+        /** @var \Psr\EventDispatcher\EventDispatcherInterface $eventDispatcherReveal */
+        $eventDispatcherReveal = $eventDispatcher->reveal();
```
itorgov, line 310: We don't need a separate variable for this value. Please get rid of this variable here and in other tests.

itorgov (changes requested):

packages/EasyCore/tests/Doctrine/Dispatchers/DeferredEntityEventDispatcherTest.php
mvlad: Currently, we need it to avoid errors from phpstran:

```
 -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  packages/EasyCore/tests/Doctrine/Dispatchers/DeferredEntityEventDispatcherTest.php:255
 -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  - '#Parameter \#1 \$eventDispatcher of class EonX\\EasyCore\\Doctrine\\Dispatchers\\DeferredEntityEventDispatcher constructor expects Psr\\EventDispatcher\\EventDispatcherInterface, object given#'
 -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Can we fix it in a separate PR for all tests?

itorgov (approved):

roman-eonx (approved):

PR #741, opened 2021-09-17
[EasyTest] Add new invalid data generators

alexndlm:
Add new invalid data providers:
- yieldInvalidAlphanumeric
- yieldInvalidAlphanumericHyphen
- yieldInvalidDateInterval
- yieldInvalidFloat

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


packages/EasyTest/src/InvalidDataMaker/InvalidDataMaker.php
```diff
@@ -6,6 +6,10 @@
 
 use Carbon\Carbon;
 use Doctrine\Common\Collections\ArrayCollection;
+use EonX\EasyCore\Bridge\Symfony\Validator\Constraints\Alphanumeric;
```
roman-eonx, line 9: We need to add `EasyCore` to the dependencies in `composer.json`.

roman-eonx (changes requested):

packages/EasyTest/src/InvalidDataMaker/InvalidDataMaker.php
alexndlm: Done.

PR #756, opened 2021-09-22
[EasyTest] Add coverage resolver for *.clover format.

voodooism:
Command easy-test check-coverage in eonx-com/easy-test show only success or error message when coverage percent is lower than expected.
As soon as the text format doesn’t show any information about uncovered files, I think it would be a good idea to support *.clover format.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no    
| Deprecations? | no
| Tests pass?   | yes


alexndlm (commented):
Good gob!

alexndlm (changes requested):
Add please test case when some file has coverage 0%, and other two files have coverage 100%

voodooism:
> Add please test case when some file has coverage 0%, and other two files have coverage 100%

And what do you want to test there? Correctness of counting common coverage? Or showing violations?
I think we can extend one of the existing test cases, but I need to understand your idea

alexndlm:
> > Add please test case when some file has coverage 0%, and other two files have coverage 100%
> 
> @alexndlm And what do you want to test there? Correctness of counting common coverage? Or showing violations?
> I think we can extend one of the existing test cases, but I need to understand your idea

I am just want to be sure that the command will correctly show files that have no coverage.

voodooism:
> > > Add please test case when some file has coverage 0%, and other two files have coverage 100%
> > 
> > 
> > @alexndlm And what do you want to test there? Correctness of counting common coverage? Or showing violations?
> > I think we can extend one of the existing test cases, but I need to understand your idea
> 
> I am just want to be sure that the command will correctly show files that have no coverage.

I've added an uncovered file in the fixture report. I personally don't think that we need to write a new test cuz [this one](https://github.com/eonx-com/easy-monorepo/blob/feature/clover-coverage-resolver/packages/EasyTest/tests/CheckCoverageCommandTest.php#L43) is covering that case

alexndlm (dismissed):

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
```diff
@@ -0,0 +1,35 @@
...
+
+use EonX\EasyTest\Coverage\Resolvers\CloverCoverageResolver;
+use EonX\EasyTest\Coverage\Resolvers\TextCoverageResolver;
+use EonX\EasyTest\Interfaces\CoverageResolverFactoryInterface;
+use EonX\EasyTest\Interfaces\CoverageResolverInterface;
+use InvalidArgumentException;
+
+final class CoverageResolverFactory implements CoverageResolverFactoryInterface
+{
+    private const CLOVER_EXTENSION = 'clover';
```
akl-95, line 15: In most cases we name constants with the rule "the more generic the word the closer to the beginnig".
It allows us to sort constants "by groups". Pls follow this rule here too and name it `EXTENSION_CLOVER`. The same for another const.

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
```diff
@@ -0,0 +1,35 @@
...
+
+    private const TXT_EXTENSION = 'txt';
+
+    public function create(string $filePath): CoverageResolverInterface
+    {
+        $ext = pathinfo($filePath, PATHINFO_EXTENSION);
```
akl-95, line 21: Why not full `$extension`? It would be more readable with no impact as I see.

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyTest\Coverage\Factory;
```
akl-95, line 5: Pls make it plural to be consistent with other namespaces.

packages/EasyTest/src/Coverage/Resolvers/CloverCoverageResolver.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        $totalElements = (int)$xml->xpath('//project/metrics/@elements')[0];
+        $totalCoveredElements = (int)$xml->xpath('//project/metrics/@coveredelements')[0];
+        $coverage = ($totalCoveredElements / $totalElements) * 100;
+
+        return new CoverageReportDto($coverage, $violations);
+    }
+
+    private function extractXMLAttribute(SimpleXMLElement $element, string $attributeName): string
```
akl-95, line 62: Why not in CamelCase? I.e. `extractXmlAttribute`

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
```diff
@@ -0,0 +1,35 @@
...
+    private const CLOVER_EXTENSION = 'clover';
+
+    private const TXT_EXTENSION = 'txt';
+
+    public function create(string $filePath): CoverageResolverInterface
+    {
+        $ext = pathinfo($filePath, PATHINFO_EXTENSION);
+        switch ($ext) {
+            case self::CLOVER_EXTENSION:
+                $resolver = new CloverCoverageResolver();
```
akl-95, line 24: I believe we must use these resolvers as services. I.e. to fetch them from the container. What about to use the same solution as we use in Payments in Locator-classes instead of factories?

packages/EasyTest/tests/CheckCoverageCommandTest.php
```diff
@@ -40,26 +64,33 @@ public function providerCheckCoverage(): iterable
...
 
-        yield 'File and Lines but no coverage' => [
+        yield 'Txt file and Lines but no coverage' => [
             [
                 'file' => __DIR__ . '/fixtures/lines-but-no-coverage.txt',
             ],
             UnableToResolveCoverageException::class,
         ];
+
+        yield 'Clover file but wrong format' => [
```
akl-95, line 88: What about a "not found" case for Clover? Or about to not split the "not found" case for Clover and Txt?

akl-95 (changes requested):

packages/EasyTest/tests/CheckCoverageCommandTest.php
voodooism: It does not matter for [resolver](https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasyTest/src/Coverage/Loaders/FilesystemCoverageLoader.php#L18) which extension nonexistent file has, so I decided not to add such case.

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
voodooism: As soon as I've changed a factory to a locator it does not make sense

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
voodooism: It's Locators now

packages/EasyTest/src/Coverage/Resolvers/CloverCoverageResolver.php
voodooism: done

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
voodooism: done

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
voodooism: done

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
akl-95: And what about the DataTransferObject namespace?

packages/EasyTest/tests/CheckCoverageCommandTest.php
akl-95: Yep, I agree, And that's why I'm talking about to not split these cases. I.e. we can name the first key as `File not found` instead of `Txt file not found`. It would be more clear I believe.

roman-eonx (dismissed):

packages/EasyTest/tests/CheckCoverageCommandTest.php
voodooism: Ok, I've renamed it.

packages/EasyTest/src/Coverage/Factory/CoverageResolverFactory.php
voodooism: done

roman-eonx (dismissed):

akl-95 (dismissed):

alexndlm (approved):

packages/EasyTest/src/Coverage/Locators/CoverageResolverLocator.php
```diff
@@ -0,0 +1,34 @@
...
+    private $serviceLocator;
+
+    public function __construct(ServiceProviderInterface $serviceLocator)
+    {
+        $this->serviceLocator = $serviceLocator;
+    }
+
+    public function getCoverageResolver(string $filePath): CoverageResolverInterface
+    {
+        $extension = pathinfo($filePath, PATHINFO_EXTENSION);
```
natepage, line 26: Please namespace the `\pathinfo()` method.

natepage: Please namespace the `\PATHINFO_EXTENSION` constant as well.

natepage (changes requested):

packages/EasyTest/src/Coverage/Locators/CoverageResolverLocator.php
voodooism: done

natepage (approved):

PR #771, opened 2021-09-30
[EasyCore] Add change sets to events dispatched by `DeferredEntityEventDispatcher`

mvlad:
We need to know what changes were made after an entity was updated. To be consistent we going to do the same for entity creations and deletions.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes


mvlad:
@roman-eonx @itorgov please see an alternative implementation - https://github.com/eonx-com/easy-monorepo/commit/b765f5b8185ac1529ad38592fd3c2614a7dbf89e

roman-eonx:
> @roman-eonx @itorgov please see an alternative implementation - [b765f5b](https://github.com/eonx-com/easy-monorepo/commit/b765f5b8185ac1529ad38592fd3c2614a7dbf89e)

Looks good. But I'd keep the method names. The initial idea of this naming was: we "defer" first and "dispatch" at the end.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -36,9 +41,28 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
         $this->eventDispatcher = $eventDispatcher;
     }
 
+    /**
+     * @param int $transactionNestingLevel
```
itorgov, line 45: You have defined these params in the interface. Do we really need to define them twice?

packages/EasyCore/src/Doctrine/Subscribers/EntityEventSubscriber.php
```diff
@@ -48,11 +57,19 @@ public function onFlush(OnFlushEventArgs $eventArgs): void
         $scheduledEntityUpdates = $this->filterEntities($unitOfWork->getScheduledEntityUpdates());
```
itorgov, line 57: You have forgot to call `computeChangeSets`.

packages/EasyCore/tests/Doctrine/Dispatchers/DeferredEntityEventDispatcherTest.php
```diff
@@ -1,359 +0,0 @@
...
-use EonX\EasyCore\Doctrine\Events\EntityUpdatedEvent;
-use EonX\EasyCore\Tests\AbstractTestCase;
-use Prophecy\Argument;
-use Psr\EventDispatcher\EventDispatcherInterface;
-use stdClass;
-
-/**
- * @covers \EonX\EasyCore\Doctrine\Dispatchers\DeferredEntityEventDispatcher
- */
-final class DeferredEntityEventDispatcherTest extends AbstractTestCase
```
itorgov, line 18: Why did you remove it?

itorgov (changes requested):

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: These params have been removed along with the method.

packages/EasyCore/src/Doctrine/Subscribers/EntityEventSubscriber.php
mvlad: Why do we need the call here?
As I see, it's already called in `\Doctrine\ORM\UnitOfWork::commit`:
1. Dispatch `preFlush`
2. `computeChangeSets`
3. Dispatch `onFlush` (out logic is here)
4. Begin and commit transaction

packages/EasyCore/tests/Doctrine/Dispatchers/DeferredEntityEventDispatcherTest.php
mvlad: I did it for the following reasons:
1. Now `EntityEventSubscriberTest` is the integration test that covers all cases that were in `DeferredEntityEventDispatcherTest`.
2. It's difficult to maintain `DeferredEntityEventDispatcherTest` because there are a lot of implementation details.

But if you think that `DeferredEntityEventDispatcherTest` can catch some errors, I will return it.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -39,90 +44,123 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
...
+        if (isset($this->entityInsertions[$oid]) !== false) {
+            return new EntityCreatedEvent($this->entityInsertions[$oid], $entityChangeSet);
+        }
+
+        if (isset($this->entityUpdates[$oid]) !== false) {
+            return new EntityUpdatedEvent($this->entityUpdates[$oid], $entityChangeSet);
+        }
+
+        // @codeCoverageIgnoreStart
+        throw new \LogicException('Entity fot event not found');
```
roman-eonx, line 144: Please fix the typo and add a point at the end of the message.

roman-eonx (changes requested):

packages/EasyCore/src/Doctrine/Subscribers/EntityEventSubscriber.php
itorgov: Ok

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -39,90 +44,123 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
...
-        $this->entityUpdates[$transactionNestingLevel] = $mergedEntityUpdates;
+        $oid = \spl_object_hash($object);
+        $this->entityUpdates[$oid] = $object;
+        $this->entityChangeSets[$transactionNestingLevel][$oid] = $entityChangeSet;
     }
 
     public function disable(): void
     {
         $this->enabled = false;
+        $this->clear();
```
itorgov, line 92: We don't have to clear it during disabling. Just imagine the following flow:

- Start a transaction.
- Create the entity A.
- Disable the dispatcher.
- Update the entity B.
- Enable the dispatcher.
- Commit the transaction.

This way we should dispatch an event about the entity A.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -39,90 +44,123 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
...
-                }
-                $processedEntities[$oid] = $entity;
-                $this->eventDispatcher->dispatch(new EntityUpdatedEvent($entity));
+            foreach ($mergedEntityChangeSets as $oid => $entityChangeSet) {
+                $event = $this->createEntityEvent($oid, $entityChangeSet);
+
+                $this->eventDispatcher->dispatch($event);
             }
+        } finally {
+            $this->clear();
```
itorgov, line 118: Why does this approach better then the previous one?

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -39,90 +44,123 @@ public function __construct(EventDispatcherInterface $eventDispatcher)
     public function clear(?int $transactionNestingLevel = null): void
     {
         if ($transactionNestingLevel !== null) {
```
itorgov, line 46: Why don't you clear the `$this->entityInsertions` and `$this->entityUpdates` arrays for the specified transaction level?

itorgov (changes requested):

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: 1. I need `entityInsertions` and `entityUpdates` in a separate method `createEntityEvent`. In the previous approach, these fields are empty.
2. This approach is a bit more optimal as it avoids a temporary copy of fields.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: 1. Now these fields don't contain transaction levels. These fields are needed only to choose the event type and construct it.
2. It's able to add transaction levels to these fields. But this will complicate the code, and I see no benefit for this (please correct me if I'm wrong).

itorgov: OK, but this way we need to add:
```
$this->entityInsertions = [];
$this->entityUpdates = [];
```
right before the early return in the `if` block.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: Fixed.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: Fixed.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: We can' add these lines before the early return in the `if` because [testEventsAreDispatchedWhenExceptionIsThrownAndCatched](https://github.com/eonx-com/easy-monorepo/blob/490ba6543df82f422bce9985659864290a68d625/packages/EasyCore/tests/Doctrine/Subscribers/EntityEventSubscriberTest.php#L68) will not pass.
Wrote a logic to clean up `entityInsertions`/`entityUpdates`.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -93,36 +112,73 @@ public function disable(): void
...
+        if (isset($this->entityInsertions[$oid]) !== false) {
+            return new EntityCreatedEvent($this->entityInsertions[$oid], $entityChangeSet);
+        }
+
+        if (isset($this->entityUpdates[$oid]) !== false) {
+            return new EntityUpdatedEvent($this->entityUpdates[$oid], $entityChangeSet);
+        }
+
+        // @codeCoverageIgnoreStart
+        throw new \LogicException('Entity for event not found.');
```
roman-eonx, line 162: This class needs to be imported.

roman-eonx (changes requested):
Let's rework this PR by introducing a new `EasyDoctrine` package and copying the contents of `easy-core/src/Doctrine` to it:
- Move `FixPostgreSqlDefaultSchemaListener` from `Events` to `Listeners`.
- Don't copy `ConnectionChecker`.
- Deprecate the original classes (similar to how it's done [here](https://github.com/eonx-com/easy-core/blob/master/src/Doctrine/DBAL/ConnectionChecker.php#L10)).
- Apply the changes from this PR only to the classes of the new package.
- Update the PR description accordingly.

packages/EasyCore/src/Doctrine/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: Fixed in `EasyDoctrine` package - https://github.com/eonx-com/easy-monorepo/blob/33c75a99d89874fa21b7e1e78c8fca85c6a37332/packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php#L9

mvlad:
New PR - https://github.com/eonx-com/easy-monorepo/pull/775

PR #798, opened 2021-11-01
[EasyDoctrine] Add EntityDeletedEvent

mvlad:
We have the `EonX\EasyDoctrine\Events\EntityCreatedEvent` and `EonX\EasyDoctrine\Events\EntityUpdatedEvent` events. Now we also need to know about deleted entities.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php
+++ b/packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php
@@ -95,7 +95,8 @@ final class DeferredEntityEventDispatcher implements DeferredEntityEventDispatch
         }
 
         $oid = \spl_object_hash($object);
-        $this->entityDeletions[$oid] = $object;
+        // `clone` is used to preserve the identifier that is removed after deleting entity
+        $this->entityDeletions[$oid] = clone $object;
         $this->entityChangeSets[$transactionNestingLevel][$oid] = $entityChangeSet;
     }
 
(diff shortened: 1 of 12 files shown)


packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -147,7 +168,7 @@ public function enable(): void
      * @param string $oid
      * @param array<string, array{mixed, mixed}> $entityChangeSet
      *
-     * @return \EonX\EasyDoctrine\Events\EntityCreatedEvent|\EonX\EasyDoctrine\Events\EntityUpdatedEvent
+     * @return \EonX\EasyDoctrine\Events\EntityCreatedEvent|\EonX\EasyDoctrine\Events\EntityUpdatedEvent|\EonX\EasyDoctrine\Events\EntityDeletedEvent
```
roman-eonx, line 171: This change makes me think we should introduce something like `EntityActionEventInterface` and change this signature.

roman-eonx (changes requested):

packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php
mvlad: Done.

roman-eonx (approved):

itorgov (approved):

natepage (approved):

PR #806, opened 2021-11-24
[EasyDoctrine] Move TimestampableImmutableTrait from AVC

skrut:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


akl-95 (changes requested):
Please move also interface and subscriber.

akl-95 (changes requested):
I think we have to change namespaces. Timestampable-related classes should be either in one `Timestampable` directory (like in AVC) or we have to split all classes by type (Trait, Interface, Subscriper).
What do you think guys?

roman-eonx:
> I think we have to change namespaces. Timestampable-related classes should be either in one `Timestampable` directory (like in AVC) or we have to split all classes by type (Trait, Interface, Subscriper). What do you think guys?

Let's keep the same structure as in other Easy-packages and introduce a specific namespace for traits and interfaces.

Please also move `EntityEventSubscriberInterface` to the new namespace.

packages/EasyDoctrine/src/Timestampable/TimestampableImmutableTrait.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDoctrine\Timestampable;
+
+use Carbon\CarbonImmutable;
```
roman-eonx, line 7: You have to add Carbon to `composer.json` this way:
```
"nesbot/carbon": "^1.39 || ^2.22"
```

roman-eonx (changes requested):

akl-95 (dismissed):

packages/EasyDoctrine/src/Timestampable/TimestampableImmutableTrait.php
skrut: done

skrut:
> 

Done, but for now EonX\\EasyDoctrine\\Subscribers\\EntityEventSubscriberInterface not found#' in `EasyActivity`

akl-95:
> Done, but for now EonX\EasyDoctrine\Subscribers\EntityEventSubscriberInterface not found#' in `EasyActivity`
Yes, but we can change the namespace in `EasyActivity` too.

skrut:
> > Done, but for now EonX\EasyDoctrine\Subscribers\EntityEventSubscriberInterface not found#' in `EasyActivity`
> 
> Yes, but we can change the namespace in `EasyActivity` too.

indeed, done

roman-eonx (approved):

akl-95 (approved):

mvlad (approved):

natepage (approved):

PR #810, opened 2021-12-02
[EasyCore] Add configuration for VirtualSearchFilter

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

When we pass any GET parameters, API platform tries to apply filters.

If `$nameConverter` is set in a filter constructor, it will be used in `\ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\AbstractFilter::denormalizePropertyName` to convert a query parameter to a string.

But to make API Platform pass any dependencies (like `$nameConverter`) into a filter while registering it, we have to define this filter as "abstract" (similar to what API Platform does in `src/vendor/api-platform/core/src/Bridge/Symfony/Bundle/Resources/config/doctrine_orm.xml`).

Otherwise, we may get the following error in some specific cases (e.g. when a query parameter name is an integer):

```
TypeError : Argument 1 passed to EonX\EasyCore\Bridge\Symfony\ApiPlatform\Filter\VirtualSearchFilter::filterProperty() must be of the type string, int given, called in /var/www/vendor/api-platform/core/src/Bridge/Doctrine/Orm/Filter/AbstractContextAwareFilter.php on line 33
 /var/www/vendor/eonx-com/easy-core/src/Bridge/Symfony/ApiPlatform/Filter/VirtualSearchFilter.php:108
 /var/www/vendor/api-platform/core/src/Bridge/Doctrine/Orm/Filter/AbstractContextAwareFilter.php:33
 /var/www/vendor/api-platform/core/src/Bridge/Doctrine/Orm/Extension/FilterExtension.php:77
```

It is recommended to always register custom filters explicitly (especially if they extend `\ApiPlatform\Core\Bridge\Doctrine\Orm\Filter\AbstractFilter`). And mark them as "abstract" if we need their dependencies to be injected by the platform.


roman-eonx (approved):

akl-95 (approved):

egor-dev (approved):

packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/filters.php
```diff
@@ -0,0 +1,31 @@
...
+use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+
+    $services->set('eonx.api_platform.doctrine.orm.virtual_search_filter', VirtualSearchFilter::class)
+        ->autoconfigure(false)
+        ->autowire(false)
+        ->args([
+            service('doctrine'),
```
natepage, line 17: `service()` function cannot be used yet as this monorepo supports Symfony 4.4
Please use `ref()` instead for now.

A seperate PR will be used to prepare the entire monorepo to either drop support for Symfony 4.4, or implement the required BC compatibility code, thanks.

natepage (changes requested):

packages/EasyCore/src/Bridge/Symfony/Resources/config/api_platform/filters.php
alexndlm: Done.

akl-95 (approved):

roman-eonx (approved):

natepage (approved):

PR #815, opened 2021-12-10
[EasyBatch] Compatibility with messenger 5.4

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

New Stamp `Symfony\Component\Messenger\Stamp\AckStamp` was added in this PR https://github.com/symfony/symfony/pull/43354.

This stamp contains closure and can't be serialized in `\EonX\EasyBatch\Transformers\AbstractBatchObjectTransformer::serialize`.

And when an exception occurs during handling messages, it causes an error - `Serialization of 'Closure' is not allowed`.

I think this Stamp isn't helpful at this point so that we can remove it before serialization.


ERuban (approved):

egor-dev (approved):

akl-95 (dismissed):

roman-eonx (dismissed):

packages/EasyBatch/src/Transformers/AbstractBatchObjectTransformer.php
```diff
@@ -86,6 +87,12 @@ protected function doTransformToArray(BatchObjectInterface $batchObject): array
 
     protected function serialize(object $message): string
     {
+        if ($message instanceof HandlerFailedException) {
```
natepage, line 90: @alexndlm nice job identifying this issue!
However this class isn't part of the Symfony Bridge for this package, so it doesn't make sense to have specific logic for it here. When implementing this serialize/unserialize logic, I thought it was simple enough to keep it in the transformer directly, but based on this issue, it might be time to extract it to its own service so we could safely fix this bug and not couple the core logic of the package with a specific framework.

Here is the way I see it working:

- Create a serializer interface for this package which defines both `serialize` and `unserialize` methods
- Create its default implementation and copy/paste the current logic from the transformer
- Inject this serializer interface into the transformer, and use it instead of the current private methods
- Create a Symfony specific implementation of the serializer interface which decorates the default one and implements your logic
- Register the default implementation as a service using a specific ID (defined as part of the bridge interface)
- Register the Symfony specific implementation to decorate the default one

What do you think?

natepage (changes requested):

packages/EasyBatch/src/Transformers/AbstractBatchObjectTransformer.php
alexndlm: Thanks for the advice!
Done.

packages/EasyBatch/src/Bridge/Symfony/Serializer/SerializerDecorator.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function __construct(SerializerInterface $decorated)
+    {
+        $this->decorated = $decorated;
+    }
+
+    public function serialize(object $message): string
+    {
+        if ($message instanceof HandlerFailedException) {
+            // @todo Use AckStamp::class when drop support Symfony lower that 5.4
```
ERuban, line 25: `that` -> `than`

ERuban (changes requested):

packages/EasyBatch/tests/Bridge/Symfony/AbstractSymfonyTestCase.php
```diff
@@ -0,0 +1,29 @@
...
+namespace EonX\EasyBatch\Tests\Bridge\Symfony;
+
+use EonX\EasyBatch\Tests\AbstractTestCase;
+use EonX\EasyBatch\Tests\Bridge\Symfony\Stubs\KernelStub;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+abstract class AbstractSymfonyTestCase extends AbstractTestCase
+{
+    /**
+     * @var \Symfony\Component\HttpKernel\KernelInterface
```
ERuban, line 14: `|null` since we have comparing with `null` in 20 line?

ERuban (changes requested):

packages/EasyBatch/tests/Bridge/Symfony/AbstractSymfonyTestCase.php
alexndlm: Done

ERuban (dismissed):

roman-eonx (approved):

packages/EasyBatch/src/Serializer/Serializer.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyBatch\Serializer;
+
+use EonX\EasyBatch\Interfaces\SerializerInterface;
+
+class Serializer implements SerializerInterface
```
akl-95, line 9: Why not final? The SerializerDecoratorTest too.

packages/EasyBatch/src/Serializer/Serializer.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyBatch\Serializer;
```
akl-95, line 5: Why not `Serializers` (plural form)?

packages/EasyBatch/src/Serializer/Serializer.php
```diff
@@ -0,0 +1,31 @@
...
+
+use EonX\EasyBatch\Interfaces\SerializerInterface;
+
+class Serializer implements SerializerInterface
+{
```
akl-95, line 10: Can't we name it more specifically? I mean `SomeSpecificSerializer` instead of generic `Serializer`.

packages/EasyBatch/src/Serializer/Serializer.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyBatch\Serializer;
+
+use EonX\EasyBatch\Interfaces\SerializerInterface;
+
```
akl-95, line 8: We don't want to test it?

packages/EasyBatch/src/Bridge/Symfony/Serializer/SerializerDecorator.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyBatch\Bridge\Symfony\Serializer;
+
+use EonX\EasyBatch\Interfaces\SerializerInterface;
+use Symfony\Component\Messenger\Exception\HandlerFailedException;
+
+class SerializerDecorator implements SerializerInterface
```
akl-95, line 10: Again, not final.

akl-95 (changes requested):

packages/EasyBatch/src/Serializer/Serializer.php
alexndlm: We use the singular form in all packages, except `EasyActivity`.
@natepage - what do you think?

akl-95: But it EasyBatch all namespaces are in plural form.

packages/EasyBatch/src/Serializer/Serializer.php
alexndlm: I think no, it contains plain logic, just a few PHP functions.

packages/EasyBatch/src/Serializer/Serializer.php
alexndlm: I see, thanks.
Dane.

packages/EasyBatch/src/Serializer/Serializer.php
alexndlm: Done.

packages/EasyBatch/src/Bridge/Symfony/Serializer/SerializerDecorator.php
alexndlm: Done.

akl-95 (approved):

egor-dev (approved):

natepage (approved):

PR #818, opened 2021-12-14
[EasyDoctrine] Solved datetime comparison issue

skrut:
- solved datetime comparison issue in easy doctrine logic

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes    
| Fixed tickets | #...

### DIFF
--- a/packages/EasyActivity/src/ActivityLogEntryFactory.php
+++ b/packages/EasyActivity/src/ActivityLogEntryFactory.php
@@ -76,7 +76,7 @@ final class ActivityLogEntryFactory implements ActivityLogEntryFactoryInterface
      */
     private function getClearedChangeSet(array $changeSet): array
     {
-        return \array_filter($changeSet, static function ($changeSetItem) {
+        return \array_filter($changeSet, static function (array $changeSetItem) {
             if ($changeSetItem[0] instanceof DateTimeInterface && $changeSetItem[1] instanceof DateTimeInterface) {
                 $format = 'Y-m-d H:i:s.uP';
 
--- a/packages/EasyActivity/src/ActivityLogEntryFactory.php
+++ b/packages/EasyActivity/src/ActivityLogEntryFactory.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace EonX\EasyActivity;
 
 use Carbon\Carbon;
-use DateTimeInterface;
 use EonX\EasyActivity\Interfaces\ActivityLogEntryFactoryInterface;
 use EonX\EasyActivity\Interfaces\ActivitySubjectDataResolverInterface;
 use EonX\EasyActivity\Interfaces\ActivitySubjectResolverInterface;
@@ -48,7 +47,6 @@ final class ActivityLogEntryFactory implements ActivityLogEntryFactoryInterface
             return null;
         }
 
-        $changeSet = $this->getClearedChangeSet($changeSet);
         $subjectData = $this->subjectDataResolver->resolve($action, $subject, $changeSet);
         if ($subjectData === null) {
             return null;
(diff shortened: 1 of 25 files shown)


packages/EasyActivity/src/ActivityLogEntryFactory.php
```diff
@@ -66,4 +68,22 @@ public function create(string $action, object $object, array $changeSet): ?Activ
...
+    /**
+     * @param mixed[] $changeSet
+     *
+     * @return mixed[]
+     */
+    private function getClearedChangeSet(array $changeSet): array
+    {
+        return \array_filter($changeSet, static function (array $changeSetItem) {
+            if ($changeSetItem[0] instanceof DateTimeInterface && $changeSetItem[1] instanceof DateTimeInterface) {
+                $format = 'Y-m-d H:i:s.uP';
```
akl-95, line 81: Let's create `DATETIME_COMPARISON_FORMAT` for it.

akl-95 (changes requested):

packages/EasyActivity/src/ActivityLogEntryFactory.php
roman-eonx: Can't we use Carbon to compare two dates instead of formatting them?

packages/EasyActivity/src/ActivityLogEntryFactory.php
```diff
@@ -47,6 +48,7 @@ public function create(string $action, object $object, array $changeSet): ?Activ
             return null;
         }
 
+        $changeSet = $this->getClearedChangeSet($changeSet);
```
roman-eonx, line 51: I think this issue should be solved not in EasyActivity but in EasyDoctrine. Here we rely on events dispatched by EasyDoctrine. And they shouldn't have DateTime objects that weren't changed in event changesets.

This issue is actual for any code relying on EasyDoctrine events. For example, webhooks.

@itorgov what do you think?

roman-eonx (changes requested):

packages/EasyActivity/src/ActivityLogEntryFactory.php
itorgov: How exactly do you suppose to compare them with Carbon?

packages/EasyActivity/src/ActivityLogEntryFactory.php
itorgov: You know, you are right. Right now `EasyDoctrine` will send an entity updated event when we update entity with the same value in any `date` attribute. It's wrong behavior. And the bad thing that Doctrine will send an update request to a database anyway :-(

itorgov (changes requested):

packages/EasyActivity/src/ActivityLogEntryFactory.php
roman-eonx: `Carbon::equalTo`, no?

skrut: For `Carbon::equalTo` dates {date = "2021-12-20 06:38:59.446547", timezone_type = {int} 3, timezone = "UTC"} {date = "2021-12-20 13:38:59.446547", timezone_type = {int} 3, timezone = "Asia/Krasnoyarsk"}  are equal. It was decided to keep the changeSet if timezone was changed.

skrut: > `Carbon::equalTo`, no?

For `Carbon::equalTo` dates {date = "2021-12-20 06:38:59.446547", timezone_type = {int} 3, timezone = "UTC"} {date = "2021-12-20 13:38:59.446547", timezone_type = {int} 3, timezone = "Asia/Krasnoyarsk"}  are equal. It was decided to keep the changeSet if timezone was changed.

packages/EasyDoctrine/tests/Subscribers/EntityEventSubscriberTest.php
```diff
@@ -20,6 +22,75 @@
...
+                $category,
+                [
+                    'activeTill' => [$activeTill, $newActiveTill],
+                ]
+            ),
+            $events[0]
+        );
+    }
+
+    public function testEventIsNotDispatchedForEqualsDates(): void
```
roman-eonx, line 65: `Equals` -> `Equal`.

roman-eonx (changes requested):

packages/EasyDoctrine/tests/Subscribers/EntityEventSubscriberTest.php
skrut: Done

akl-95 (approved):

roman-eonx (approved):

itorgov (approved):

PR #809, opened 2021-12-02
Updated Symfony configs

itorgov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 

- Updated Symfony configs in each package.
- [EasyDoctrine] Type-hinting improvements.
- [EasySecurity] Improved the `ContextAuthenticator` class.
- [EasyCore] Added the `getCommandDescription` method to the `ApiResourceAndSimpleDataPersisterMaker` command.
- [EasyCore] Updated the constructor in the `ResolveRequestAttributesListener` listener to fix a deprecation notice.
- [EasyRequestId] Updated the `__invoke` method in the `RequestListener` listener to fix a deprecation notice.

PR #825, opened 2021-12-17
Replaced the `ref` function with `service` in Symfony configs for each package

itorgov:
This is the replacement for https://github.com/eonx-com/easy-monorepo/pull/809. Can be merged after we decide on how to deal with breaking changes for old Symfony versions.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


natepage:
@itorgov thank you for this PR, I will keep it on the side for now and will include it in the next major version. I'm planning on creating it over the Xmas break.

PR #829, opened 2022-01-13
[EasyActivity] Fixed potential circular reference issue

skrut:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |


packages/EasyActivity/src/Bridge/Symfony/Serializers/CircularReferenceHandlerInterface.php
```diff
@@ -0,0 +1,15 @@
...
+namespace EonX\EasyActivity\Bridge\Symfony\Serializers;
+
+interface CircularReferenceHandlerInterface
+{
+    /**
+     * @param mixed[] $context
+     *
+     * @return mixed
+     */
+    public function __invoke(object $object, string $format, array $context);
```
akl-95, line 14: We always return `string` here. What about changing the declaration to the strict return type?

akl-95 (changes requested):

packages/EasyActivity/src/Bridge/Symfony/Serializers/CircularReferenceHandler.php
```diff
@@ -0,0 +1,40 @@
...
+    public function __invoke(object $object, string $format, array $context)
+    {
+        $className = \get_class($object);
+        try {
+            $identifier = $this->entityManager->getClassMetadata($className)
+                ->getSingleIdentifierFieldName();
+            $reflectionProperty = new ReflectionProperty($className, $identifier);
+            $reflectionProperty->setAccessible(true);
+
+            return \sprintf('%s#%s (circular reference)', $className, $reflectionProperty->getValue($object));
```
mvlad, line 35: `%s#%s (circular reference)` - this value will be stored in the database.
What about returning just `$reflectionProperty->getValue($object)` by default and `null` for exceptional case? It will be easier to work with such values.

packages/EasyActivity/src/Bridge/Symfony/DependencyInjection/EasyActivityExtension.php
```diff
@@ -42,6 +45,13 @@ public function load(array $configs, ContainerBuilder $container): void
         if ($this->easyDoctrineBundleIsRegistered($container)) {
             $loader->load('easy-doctrine-bridge-services.php');
         }
+
+        $circularReferenceHandler = $config['circular_reference_handler'];
```
mvlad, line 49: Can we add the definition for `CircularReferenceHandlerInterface::class` in `packages/EasyActivity/src/Bridge/Symfony/Resources/config/services.php` as for other services?

packages/EasyActivity/tests/Stubs/CircularReferenceHandlerStub.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyActivity\Tests\Stubs;
+
+use EonX\EasyActivity\Bridge\Symfony\Serializers\CircularReferenceHandlerInterface;
+
+final class CircularReferenceHandlerStub implements CircularReferenceHandlerInterface
```
mvlad, line 9: Why do we use a Stub instead of a real service?

mvlad (changes requested):

akl-95 (changes requested):

packages/EasyActivity/src/Bridge/Symfony/Serializers/CircularReferenceHandler.php
roman-eonx: @mvlad as we just discussed with @akl-95, the idea of this new feature is to highlight a potential issue in the configs. It would be great to never store circular references in logs, so when this happens, a developer or an admin sees this in the logs and makes sure the configs are changed to exclude the corresponding relationship.

packages/EasyActivity/src/Bridge/Symfony/DependencyInjection/EasyActivityExtension.php
skrut: Circular handler is related on  application and may not be necessary default `CircularReferenceHandler`. So in `EasyActivityExtension` we get handler class from config and make it as service dynamically.

packages/EasyActivity/src/Bridge/Symfony/Serializers/CircularReferenceHandlerInterface.php
skrut: done

packages/EasyActivity/tests/Stubs/CircularReferenceHandlerStub.php
skrut: done

packages/EasyActivity/tests/Bridge/Symfony/Serializers/SymfonyActivitySubjectDataSerializerTest.php
```diff
@@ -30,6 +32,9 @@ public function provideDataForSerializeSucceeds(): iterable
         $author->setId($entityId);
         $author->setName($authorName);
         $author->setPosition($authorPosition);
+        $comment = new Comment();
```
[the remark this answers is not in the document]
skrut: done

akl-95 (approved):

akl-95 (changes requested):
Pls merge fresh `master` to fix CI.

packages/EasyActivity/tests/Bridge/Symfony/Serializers/SymfonyActivitySubjectDataSerializerTest.php
```diff
@@ -168,6 +171,23 @@ public function provideDataForSerializeSucceeds(): iterable
...
             'expectedResult' => '{"author":{"name":"John Doe"},"content":"text"}',
         ];
+
+        $comment = (new Comment())->setId(1);
+        $article = new Article();
+        $article->addComment($comment);
+        $allowedProperties = [
+            'comments' => ['article'],
+        ];
+        yield 'With circular reference' => [
```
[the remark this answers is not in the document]
skrut: Hmm, it's work for me. I added message property initialization, pls have a look again.

skrut:
> Pls merge fresh `master` to fix CI.

Done

akl-95 (dismissed):

roman-eonx (approved):

mvlad (dismissed):

packages/EasyActivity/src/Bridge/Symfony/DependencyInjection/EasyActivityExtension.php
natepage: I agree with @mvlad on this, and we already have examples in other packages so keeping things consistent would be great.
I agree with your point too @skrut , application might want to have their own CircularReferenceHandler and this will be able to do so.

- Create a unique service id for the CircularReferenceHandler of this package
- Register the default implementation in the `config/services.php` using this service id

Done! 😄 

If an application wants to customize this logic, then they simply do the same thing, create their own implementation and register it using the service id from the package, this will automatically replace the default CircularReferenceHandler with theirs.

natepage (changes requested):

packages/EasyActivity/src/Bridge/Symfony/DependencyInjection/EasyActivityExtension.php
skrut: Done

mvlad: @skrut, it seems we can remove the `circular_reference_handler` setting now. Also please replace `'easy_activity.circular_reference_handler'` with `BridgeConstantsInterface::PARAM_CIRCULAR_REFERENCE_HANDLER` in the `services.php`.

skrut: @mvlad, of course, thank you!

roman-eonx (approved):

mvlad (approved):

akl-95 (approved):

packages/EasyActivity/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -49,10 +50,13 @@
     $services
         ->alias(BridgeConstantsInterface::SERVICE_SERIALIZER, 'serializer');
 
+    $services->set(BridgeConstantsInterface::SERVICE_CIRCULAR_REFERENCE_HANDLER, CircularReferenceHandler::class);
+
     $services
         ->set(ActivitySubjectDataSerializerInterface::class, SymfonyActivitySubjectDataSerializer::class)
-        ->arg('$disallowedProperties', '%' . BridgeConstantsInterface::PARAM_DISALLOWED_PROPERTIES . '%')
-        ->arg('$serializer', service(BridgeConstantsInterface::SERVICE_SERIALIZER));
+        ->arg('$serializer', service(BridgeConstantsInterface::SERVICE_SERIALIZER))
```
natepage, line 57: If you want this PR to be merged right now, please change `service` to `ref`, then we can fix that file in the coming next version when we drop support for Symfony 4.4

natepage (changes requested):

packages/EasyActivity/src/Bridge/Symfony/Resources/config/services.php
skrut: Done, replaced by `ref`

akl-95 (approved):

roman-eonx (approved):

natepage (approved):

PR #841, opened 2022-01-28
[EasyActivity] Check if value not a date

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


packages/EasyActivity/src/Bridge/Symfony/Serializers/SymfonyActivitySubjectDataSerializer.php
```diff
@@ -71,8 +71,10 @@ public function serialize(array $data, ActivitySubjectInterface $subject): ?stri
             if (\is_object($value)) {
                 $objectClass = \get_class($value);
 
-                $context[AbstractNormalizer::ATTRIBUTES][$key] = $nestedObjectAllowedProperties[$objectClass]
-                    ?? $allowedProperties[$key] ?? ['id'];
+                if (\is_subclass_of($objectClass, \DateTimeInterface::class) === false) {
```
mvlad, line 74: It seems we are trying to avoid adding `'id'` to the context because it may not be used. But is it a real problem?
There are a few possible cases:
* `DateTimeNormalizer` is added to `Serializer`. In this case, `AbstractNormalizer::ATTRIBUTES` will be ignored.
* `ObjectNormalizer` is added to `Serializer`. In this case, `id` in `AbstractNormalizer::ATTRIBUTES` will be ignored and an empty object will be returned.
* Normalizers are not defined. An exception will occur.

So, I don't see any issues with adding `id` to the context. And perhaps we don't have to complicate the code.

If we anyway want to keep the context clean, what about relying on `$id` property? For example, we can check it using `symfony/property-access`.

mvlad (changes requested):

packages/EasyActivity/src/Bridge/Symfony/Serializers/SymfonyActivitySubjectDataSerializer.php
akl-95: The initial reason for this fix was an incorrect context. I'm talking about "incorrect" because according to this context we require the `id` property on objects which may not have this `id` at all. It's not a critical problem, you are right.
But it seemed a little strange to me when debugging.

So my point is that we shouldn't "require" non-existent attributes, even if it doesn't cause issues. But also I believe you'll make a more correct decision about this since you know this feature better than us.

I believe we shouldn't use `is_subclass_of` here and should use `symfony/property-access` as you said.

akl-95 (changes requested):
Pls provide fixes according to the discussion above.

packages/EasyActivity/src/Bridge/Symfony/Serializers/SymfonyActivitySubjectDataSerializer.php
alexndlm: When using `symfony/property-access` it couse to many edge cases, as `$value` may be `ArrayCollection` ...

I prefer to use `is_subclass_of` or just close this PR for now.

alexndlm:
We decided to leave as it is because we can't find a good solution.
This bug doesn't create any real problems.

PR #843, opened 2022-01-28
[EasyDoctrine] Use myclabs/deep-copy

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


mvlad (approved):

packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php
```diff
@@ -95,8 +98,14 @@ public function deferDelete(int $transactionNestingLevel, object $object, array
         }
 
         $oid = \spl_object_hash($object);
-        // `clone` is used to preserve the identifier that is removed after deleting entity
-        $this->entityDeletions[$oid] = clone $object;
+        // DeepCopy is used to preserve the identifier that is removed after deleting entity
+        $copier = new DeepCopy();
```
akl-95, line 102: Let's inject it as a service.

akl-95 (changes requested):

packages/EasyDoctrine/src/Dispatchers/DeferredEntityEventDispatcher.php
alexndlm: Have done.

PR #845, opened 2022-02-03
[EasyErrorHandler] Move AnalyzeErrorCodesCommand to EasyErrorHandler

DKeeper:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


alexndlm (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Command/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Command;
```
akl-95, line 5: Pls change namespace to `Commands` to be consistent.

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/ErrorCodesProviderInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Interfaces;
+
+interface ErrorCodesProviderInterface
+{
+    /**
+     * @return mixed[]
```
akl-95, line 10: Can't we write `int[]`?

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/ErrorCodesProviderInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Interfaces;
```
akl-95, line 5: I believe we should place it under `EonX\EasyErrorHandler\Bridge\Symfony\Interfaces\ErrorCodes`. Wdyt?

akl-95 (changes requested):

voodooism (dismissed):

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/ErrorCodesProviderInterface.php
DKeeper: Agree, Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/ErrorCodesProviderInterface.php
DKeeper: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Command/AnalyzeErrorCodesCommand.php
DKeeper: Done.

DKeeper: Done.

akl-95 (dismissed):

alexndlm (dismissed):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,126 @@
...
+{
+    /**
+     * @var int
+     */
+    private const DEFAULT_CATEGORY_SIZE = 100;
+
+    /**
+     * @var string
+     */
+    protected static $defaultName = 'error-codes:analyze';
```
itorgov, line 23: What about to add a prefix here like doctrine does? Something like `eonx:error-codes:analyze` or `easy-error-handler:error-codes:analyze`.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,126 @@
...
+use EonX\EasyErrorHandler\Bridge\Symfony\Interfaces\ErrorCodes\ErrorCodesProviderInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class AnalyzeErrorCodesCommand extends Command
+{
+    /**
+     * @var int
```
itorgov, line 16: As I know @natepage agreed that we can omit annotations for constants.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,126 @@
...
+     */
+    private $categorySize;
+
+    /**
+     * @var \EonX\EasyErrorHandler\Bridge\Symfony\Interfaces\ErrorCodes\ErrorCodesProviderInterface
+     */
+    private $errorCodesProvider;
+
+    public function __construct(ErrorCodesProviderInterface $errorCodesProvider, ?int $categorySize = null)
```
itorgov, line 35: Please use the property promotion feature here.

itorgov (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
voodooism: Only PHP 8+ supports the feature, doesn't it?

voodooism (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
roman-eonx: @voodooism `master` is v4 which is restricted to PHP 8.1+ :wink:

voodooism: In that case, I think we can use strict types too in this PR ?

voodooism (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
akl-95: I vote for `easy-error-handler:error-codes:analyze`

DKeeper: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
DKeeper: Refactored.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
DKeeper: Refactored.

alexndlm (approved):

akl-95 (dismissed):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,110 @@
...
+final class AnalyzeErrorCodesCommand extends Command
+{
+    /**
+     * @var string
+     */
+    protected static $defaultName = 'easy-error-handler:error-codes:analyze';
+
+    public function __construct(
+        private ErrorCodesProviderInterface $errorCodesProvider,
+        private int $categorySize = 100
```
itorgov, line 22: Sorry but I think we shouldn't use a default value here because of company's code style rules.
I should be `private ?int $categorySize = null` and then in the body `$this->categorySize = $categorySize ?? 100;`.

itorgov (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
itorgov: Why did you delete this constant?

voodooism (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
DKeeper: @itorgov We use the value only once, I think that using a constant to store it is unnecessary.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
DKeeper: In new projects, for such fields that have a default value, we define them in this form:
```
protected string $status = self::STATUS_ACTIVE
...
protected int $priority = 0
```

itorgov: As I know it only related to entities and properties with the `readonly` modifier.
@akl-95 wdyt?

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
itorgov: I know but usually we use constats in such cases even there is only one place where we use the value.
@akl-95 wdyt?

itorgov (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
akl-95: This const is stull useful I believe since it makes the magic value `100` clear.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
akl-95: Yes, this discussion initially appeared when working on `readonly` PHP feature. But here we have a little different case.
Let's consider what structure would we have if we don't use promoted properties:
```PHP
    private int $categorySize;

    private ErrorCodesProviderInterface $errorCodesProvider;

    public function __construct(
        ErrorCodesProviderInterface $errorCodesProvider,
        ?int $categorySize = null
    ) {
        $this->categorySize = $categorySize ?? 100;
        parent::__construct();
    }
```

That means we have to follow one of the ways:
 - don't use promoted properties at all
 - use non-null default value for the promoted property
 
 Otherwise, we'll have the nullable property that actually is not nullable.
 So it looks pretty similar to the readonly-related issue. If we want to use promoted property here (and in other similar places) we can't use default null value, and we should update our rules.

itorgov: I just realised that I didn't use the promoted properties for such parameters in my [PR](https://github.com/eonx-com/easy-monorepo/pull/897/files#diff-19585bd0ab2a39fbfa3b4359912c3e198156561e14b4d9c4c630eb03a3c81058R34).

natepage: I'm ok with changing the standard yeah, to use non default null value so we can use promoted properties

itorgov (dismissed):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,112 @@
...
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Commands;
+
+use EonX\EasyErrorHandler\Bridge\Symfony\Interfaces\ErrorCodes\ErrorCodesProviderInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class AnalyzeErrorCodesCommand extends Command
```
natepage, line 13: I don't see any changes to the services registration, is it expected that each app will have to manually register this command?

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/ErrorCodes/ErrorCodesProviderInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Interfaces\ErrorCodes;
+
+interface ErrorCodesProviderInterface
```
natepage, line 7: No default implementation of this?

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
roman-eonx: @natepage this is a command to run manually on local machines to verify if all the error codes are defined correctly. It's not something that needs to be automated, it's just a helper for developers.

natepage: I would still register it as a command this way it's really convenient for people to run it

packages/EasyErrorHandler/src/Bridge/Symfony/ErrorCodesProvider.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Bridge\Symfony;
```
alexndlm, line 5: Please move to `EonX\EasyErrorHandler\Bridge\Symfony\Providers`

alexndlm (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/ErrorCodesProvider.php
DKeeper: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Interfaces/ErrorCodes/ErrorCodesProviderInterface.php
DKeeper: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
DKeeper: Done.

alexndlm (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesProvider.php
```diff
@@ -0,0 +1,21 @@
...
+final class ErrorCodesProvider implements ErrorCodesProviderInterface
+{
+    /**
+     * @return array<string, int>
+     *
+     * @throws \EonX\EasyErrorHandler\Exceptions\ErrorCodesProviderException
+     */
+    public function provide(): array
+    {
+        throw new ErrorCodesProviderException('exceptions.error_codes_provider.not_configured');
```
akl-95, line 19: The current provider implementation makes no sense I believe. And it makes us think about another way to work with this feature. Let's try the following way:
 - introduce `ErrorCodesInterface` in this package and extend it by the app-level `ErrorCodesInterface`
 - keep `ErrorCodesProviderInterface` and implement `ErrorCodesProvider` with the full providing logic, that we have in Payments now, using the `ErrorCodesInterface` from the point above

That will allow us to have a working default implementation of the provider based on the package-level `ErrorCodesInterface`. All we need to do on the app-level then is only to extend `ErrorCodesInterface`.

akl-95 (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesProvider.php
```diff
@@ -0,0 +1,40 @@
...
+use Symfony\Component\ErrorHandler\Error\ClassNotFoundError;
+
+final class ErrorCodesProvider implements ErrorCodesProviderInterface
+{
+    private const ERROR_CODE_NAME_PREFIX = 'ERROR_';
+
+    /**
+     * @param class-string $errorCodesInterface
+     */
+    public function __construct(private string $errorCodesInterface)
```
alexndlm, line 19: Let's pass here a service that implements `ErrorCodesInterface`.
In the app, we can redefine this service.

I think it is better than binding variable `$errorCodesInterface`.

alexndlm (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesProvider.php
roman-eonx: We won't implement this interface in a service. The idea is to have `ErrorCodesInterface extends <EasyErrorHandler>ErrorCodesInterface` in the app. But I agree that binding this variable explicitly is not a good idea.

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/translations/EasyErrorHandlerBundle.en.php
```diff
@@ -8,6 +8,9 @@
         'bad_request' => 'Bad request.',
         'conflict' => 'Conflict.',
         'entity_not_valid' => 'Entity validation failed.',
+        'error_codes_provider' => [
```
akl-95, line 11: Pls remove it, it's not relevant anymore.

akl-95 (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesProvider.php
akl-95: But what can we do other than binding?

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/translations/EasyErrorHandlerBundle.en.php
DKeeper: Done.

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesProvider.php
akl-95: @alexndlm that means I would need to define `error_codes_interface` even if I don't want this feature.

alexndlm: @akl-95 hmmmm, yes.

packages/EasyErrorHandler/src/Bridge/Symfony/Commands/AnalyzeErrorCodesCommand.php
```diff
@@ -0,0 +1,118 @@
...
+{
+    private const DEFAULT_CATEGORY_SIZE = 100;
+
+    /**
+     * @var string
+     */
+    protected static $defaultName = 'easy-error-handler:error-codes:analyze';
```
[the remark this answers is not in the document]
DKeeper: Done.

akl-95 (dismissed):

roman-eonx (dismissed):

itorgov (dismissed):

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -26,7 +30,8 @@
     $services = $containerConfigurator->services();
     $services->defaults()
         ->autowire()
-        ->autoconfigure();
+        ->autoconfigure()
+        ->bind('string $errorCodesInterface', ErrorCodesInterface::class);
```
natepage, line 34: I'm not 100% sure what's the objective with this and the default ErrorCodesProvider but `bind` is working only within the context of the current file which means apps won't be able to override it from their own config files, and it feels strange to create an empty interface to use reflection on it.

Could you please clarify what is the intention with this default error codes provider?

- For apps to use it by changing the class passed to its constructor? Then I think we shouldn't have an empty interface but have a nullable config to allow apps to define which class to use, and if it's null then the provider simply returns an empty array.
- For apps to replace the error codes provider with their own? Then let's simply return an empty array from the default one and it's done.

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
akl-95: The first option, i.e. to allow apps to define which class to use as error codes source. Yes, we can use nullable `$errorCodesInterface` to get rid of this empty `ErrorCodesInterface`.

natepage: @akl-95 ok then I think we need to introduce a config for it

DKeeper: Reworked.

akl-95 (approved):

roman-eonx (dismissed):

akl-95:
We're gonna rework it and create a new one.

PR #866, opened 2022-03-10
[EasyCore] hotfix/easy-core-decimal-validator

alexandrtspl:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    

In the decimalValidator, the value being tested can be converted to an exponential form. And it not will match the regexp pattern.
Eg: "0.000001" to "1.0E-6"

This one should fix that.

### DIFF
--- a/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/DecimalValidator.php
+++ b/packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/DecimalValidator.php
@@ -26,7 +26,9 @@ final class DecimalValidator extends ConstraintValidator
             throw new UnexpectedValueException($value, 'scalar');
         }
 
-        $value = (string) $value;
+        $value = \is_float($value)
+            ? \rtrim(\number_format($value, $constraint->maxPrecision + 1, '.', ''), '.0')
+            : (string) $value;
 
         $pattern = \sprintf('/^\d+(\.\d{%d,%d})?$/', $constraint->minPrecision, $constraint->maxPrecision);
 
--- a/packages/EasyCore/tests/Bridge/Symfony/Validator/Constraints/DecimalValidatorTest.php
+++ b/packages/EasyCore/tests/Bridge/Symfony/Validator/Constraints/DecimalValidatorTest.php
@@ -50,6 +50,7 @@ final class DecimalValidatorTest extends AbstractSymfonyTestCase
             'Valid value #5' => [0.01, 1, 3],
             'Valid value #6' => [1, 1, 3],
             'Valid value #7' => [0, 1, 3],
+            'Valid value #8' => [0.000001, 1, 6],
             'Empty string' => ['', 1, 3],
             'Null value' => [null, 1, 3],
         ];


zeromodule (approved):

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/DecimalValidator.php
```diff
@@ -26,7 +26,9 @@ public function validate($value, Constraint $constraint): void
             throw new UnexpectedValueException($value, 'scalar');
         }
 
-        $value = (string) $value;
+        $value = \is_float($value)
+            ? \rtrim(\number_format($value, $constraint->maxPrecision + 1, '.', ''), '.0')
```
akl-95, line 30: Why do we need precision + 1 and `rtrim`?

akl-95 (changes requested):

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/DecimalValidator.php
alexandrtspl: For number_format we need precision (it's required). If we will use $constraint->maxPrecision our value after number_format will always be correct even it was bigger. For example: 

Why we need `Precision + 1`:

```
$value = 0.000001234  
maxPrecision = 6
```

`$value = number_format(0.000001234, 6, '.', '');`
`$value = "0.000001" `it's always correct value. with precision+1 we can get incorrect value to later check it.

`$value = number_format(0.000001234, 6+1, '.', '');`
`$value = "0.0000012"`  - it's incorrect and we check it later with regexp

Why we need `Trim`:

But if our value already was correct we got a string with a trailing zero (after number_format). 
We need to trim the trailing "0" event it doesn't match regexp and it's incorrect

Example:
`$value = number_format(0.000001, 6+1)`
`$value = "0.0000010"` if we don't trim "0" we don't match with regexp.

And another situation if our input value is incorrect:

Example:
`$value = number_format(0.00000000000001, 6+1)`
`$value = "0.0000000"` if we don't trim "0" we got match with regexp.


PS. And we shouldn't trim the dot. Fixed.

akl-95: Amazing explanation, thank you. But let's consider the following case. Value is `0.0000010234`. We format it with 7 precision to  `0.0000010` and then `rtrim` turns it to 0.000001`. This value will be treated as correct but the initial one `0.0000010234` is incorrect.
Am I right?

akl-95: Can we use `sprintf` to represent the value as a string instead? In this way we shouldn't take care about trimming and about different formats. Maybe there are other issues with the `sprintf`, we need to check it.

alexandrtspl: > This value will be treated as correct but the initial one 0.0000010234 is incorrect.
> Am I right?

Yes you right. It's a bug.

> Can we use `sprintf` to represent the value as a string instead?

No we can't use `sprintf`. It will cause same problems as number_format. `sprintf`  float format require precision (or use 6 digits max by default)
```
$value = 0.0000012;
print sprintf('%.f', $value);
// string(8) "0.000001"
```
To fix this problem with sprintf or number_format we need some maximum precision and then trim it. But I throught it's ugly.

Alternative idea don't try to cast `float` to `string` but cast `string` to `float`. And change final check. 
Ex:
```
$value = (float) (string) $value;
$preparedValue = $value * 10 ** $constraint->maxPrecision;

if (($preparedValue - \floor($preparedValue)) > 0) {
    ... build violation
}
```
We need a double cast the value for objects to call `__toString` and than return it to float

Wdyt?

alexandrtspl: https://github.com/eonx-com/easy-monorepo/pull/866/commits/d299828cdff57da984e07947a0cb0551d2e93aeb

akl-95 (changes requested):
Pls provide fixes according to the conversation above.

packages/EasyCore/src/Bridge/Symfony/Validator/Constraints/DecimalValidator.php
alexandrtspl: Updated. Please check.

PR #882, opened 2022-03-28
[EasyDoctrine] Use myclabs/deep-copy

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyDoctrine/src/Bridge/Symfony/Resources/config/services.php
+++ b/packages/EasyDoctrine/src/Bridge/Symfony/Resources/config/services.php
@@ -6,6 +6,9 @@ namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use EonX\EasyDoctrine\Dispatchers\DeferredEntityEventDispatcher;
 use EonX\EasyDoctrine\Dispatchers\DeferredEntityEventDispatcherInterface;
+use EonX\EasyDoctrine\Interfaces\ObjectCopierFactoryInterface;
+use EonX\EasyDoctrine\Interfaces\ObjectCopierInterface;
+use EonX\EasyDoctrine\Utils\ObjectCopierFactory;
 
 return static function (ContainerConfigurator $containerConfigurator): void {
     $services = $containerConfigurator->services();
@@ -13,6 +16,12 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $services->defaults()
         ->autowire()
         ->autoconfigure();
+
+    $services->set(ObjectCopierFactoryInterface::class, ObjectCopierFactory::class);
+
+    $services
+        ->set(ObjectCopierInterface::class)
+        ->factory([service(ObjectCopierFactoryInterface::class), 'create']);
 
     $services->set(DeferredEntityEventDispatcherInterface::class, DeferredEntityEventDispatcher::class);
 };
--- a/packages/EasyDoctrine/src/Utils/ObjectCopier.php
+++ b/packages/EasyDoctrine/src/Utils/ObjectCopier.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDoctrine\Utils;
+
+use DeepCopy\DeepCopy;
+use EonX\EasyDoctrine\Interfaces\ObjectCopierInterface;
+
+class ObjectCopier implements ObjectCopierInterface
+{
+    public function __construct(private DeepCopy $deepCopy)
+    {
+    }
+
+    public function copy(object $object): object
+    {
+        return $this->deepCopy->copy($object);
+    }
+}
--- a/packages/EasyDoctrine/src/Utils/ObjectCopierFactory.php
+++ b/packages/EasyDoctrine/src/Utils/ObjectCopierFactory.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDoctrine\Utils;
+
+use DeepCopy\DeepCopy;
+use DeepCopy\Filter\Doctrine\DoctrineCollectionFilter;
+use DeepCopy\Matcher\PropertyTypeMatcher;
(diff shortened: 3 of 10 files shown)


packages/EasyDoctrine/src/Utils/ObjectCopier.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDoctrine\Utils;
+
+use DeepCopy\DeepCopy;
+use EonX\EasyDoctrine\Interfaces\ObjectCopierInterface;
+
+class ObjectCopier implements ObjectCopierInterface
```
akl-95, line 10: What about tests for added classes?

akl-95 (changes requested):

packages/EasyDoctrine/src/Utils/ObjectCopierFactory.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyDoctrine\Utils;
+
+use DeepCopy\DeepCopy;
```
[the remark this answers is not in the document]
alexndlm: Done

packages/EasyDoctrine/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -14,5 +17,11 @@
         ->autowire()
         ->autoconfigure();
 
+    $services->set(ObjectCopierFactoryInterface::class, ObjectCopierFactory::class);
```
roman-eonx, line 20: I don't think we need an interface for this factory, as it is needed only for DI purposes. An interface for the copier itself - yes. But the factory is a Symfony way of registering the copier.

And, for the same reason, we should probably move the factory to a specific directory in `DependencyInjection`, as it is not related to the implementation itself.

@natepage wdyt?

roman-eonx (changes requested):

packages/EasyDoctrine/src/Bridge/Symfony/Resources/config/services.php
natepage: Yes I agree, no need for interface and regarding where to put the factory it's up to you guys

roman-eonx: @alexndlm let's put the factory to `EonX\EasyDoctrine\Bridge\Symfony\DependencyInjection` then and remove the interface. Probably, we don't even need to register it as a service in this line of code, as it is used as a service factory below.

alexndlm: Refactored.

roman-eonx (dismissed):

.github/workflows/php-ci.yml
```diff
@@ -1,6 +1,6 @@
 name: PHP-CI
 
-on: [ push, pull_request ]
```
roman-eonx, line 3: Why?

packages/EasyUtils/src/Math.php
```diff
@@ -10,31 +10,24 @@
...
-     * @var int
-     */
-    private $scale;
+    private int $scale;
+
+    private string $thousandsSeparator;
 
     /**
-     * @var string
+     * @param int|null $roundMode [optional] <p>
```
roman-eonx, line 24: We've never used this kind of formatting before (with `<p>`). Is this something usual for the PhpDoc standard?

roman-eonx (changes requested):

packages/EasyDoctrine/src/Utils/ObjectCopier.php
akl-95, line 10: Pls make it `final`.

akl-95 (changes requested):

.github/workflows/php-ci.yml
alexndlm: We have double checks on every new commit. We use only `push` in Payments, and all works fine.

packages/EasyUtils/src/Math.php
alexndlm: I have copied this from https://github.com/JetBrains/phpstorm-stubs/blob/385f4627a8aae0d47b5496e87638354e3bdb30d4/standard/standard_3.php#L125

And I see it is legal - https://softwareengineering.stackexchange.com/questions/157283/html-in-docblock-comments

packages/EasyDoctrine/src/Utils/ObjectCopier.php
alexndlm: Thanks. Fixed.

.github/workflows/php-ci.yml
roman-eonx: @natepage any objections?

natepage: There was a reason why I've added pull_request so let's keep it for now please

alexndlm: I have returned this.

roman-eonx (dismissed):

mvlad (changes requested):
> With a simple clone, the Product::tags links to the original collection of tags which already been removed. In that case, we do not store deleted Tag ids.

What about a test for this case?

akl-95 (dismissed):

natepage (dismissed):

alexndlm:
> > With a simple clone, the Product::tags links to the original collection of tags which already been removed. In that case, we do not store deleted Tag ids.
> 
> What about a test for this case?

finally-done-meme

mvlad (approved):

roman-eonx (approved):

akl-95 (approved):

natepage (approved):

PR #884, opened 2022-03-29
[EasyDoctrine] Rename deferred events

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes

Events classes are renamed not to be confused about the source of events.


roman-eonx:
From the package consumer perspective, these events are named according to their purposes: the entity being created/updated/deleted. "Deferred" is more about the implementation details of the package.

So, if we don't have some real reasons for changing these names, I'd prefer not to introduce this breaking change (that will require a bunch of changes in the projects after updating).

alexndlm:
> From the package consumer perspective, these events are named according to their purposes: the entity being created/updated/deleted. "Deferred" is more about the implementation details of the package.
> 
> So, if we don't have some real reasons for changing these names, I'd prefer not to introduce this breaking change (that will require a bunch of changes in the projects after updating).

With current naming developers must always remember that events and handlers executed deferred, and remembers to set this in configuration.

Proof - https://github.com/eonx-com/payment-gateway-api-v3/pull/1634/files. Webhooks for OrderIntent works before just because of the EasyActivity config.

roman-eonx:
> With current naming developers must always remember that events and handlers executed deferred, and remembers to set this in configuration.

Developers must remember that the entities need to be included in the config explicitly, yes. But this is not about deferring those events, it's about how the package works. I understand that you are talking about "entity created" in the event name may confuse a developer thinking this event will be dispatched automatically, without an explicit config. But even if we rename the events, the situation won't change much. Developers must know and remember about this package feature.

alexndlm:
What about deprecating auto prepending EasyDoctrine configs?

alexndlm:
Now EasyActivity automatically adds EasyDoctrine configs.

Please see https://github.com/eonx-com/easy-monorepo/blob/fc39195f7feb931b8dc2ea4288c695dc3d55917c/packages/EasyActivity/src/Bridge/Symfony/DependencyInjection/EasyActivityExtension.php#L47

This makes application configs less explicitly.

roman-eonx:
How will this help developers not forget about configs? This will make the situation even worse, as they will need to remember to add every new entity to 2 configs instead of 1.

alexndlm:
This is unlikely to help, but it will make app configs explicit. I say so because recently spend some time finding out how Webhooks works without configs.

roman-eonx:
I don't think that adding an entity to 2 configs is better than adding it to 1 (so the package automatically "copies" it to another one). The chances of forgetting to add it to 2 configs are higher than the chances of forgetting to add it to 1 config, right?

PR #903, opened 2022-04-25
[EasyActivity] Use Symfony UuidFactory to generate IDs for ActivityLog

mvlad:
Currently we use `\EonX\EasyRandom\RandomGenerator::uuidV4()` to generate IDs for `ActivityLog`.
It makes sense to use `Symfony\Component\Uid\Factory\UuidFactory` for the following reasons:
* consistency with other tables
* UuidV6, that Symfony uses by default, has advantages over UuidV4

How this change will affect projects:
* Projects that uses UuidV4 (has a `default_uuid_version = 4` setting) will not be affected
* Projects that uses UuidV6 will save activity logs with a new ID format. But as UuidV4  is randomly generated, this change should not break anything.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyActivity/src/Bridge/Symfony/Uid/UuidFactory.php
+++ b/packages/EasyActivity/src/Bridge/Symfony/Uid/UuidFactory.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyActivity\Bridge\Symfony\Uid;
+
+use EonX\EasyActivity\Interfaces\IdFactoryInterface;
+use Symfony\Component\Uid\Factory\UuidFactory as SymfonyUuidFactory;
+
+final class UuidFactory implements IdFactoryInterface
+{
+    public function __construct(private SymfonyUuidFactory $uuidFactory)
+    {
+        // The body is not required
+    }
+
+    public function create(): int|string
+    {
+        return (string)$this->uuidFactory->create();
+    }
+}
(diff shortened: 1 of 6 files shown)


packages/EasyActivity/src/Bridge/Symfony/Uid/UuidFactory.php
```diff
@@ -0,0 +1,21 @@
...
+use Symfony\Component\Uid\Factory\UuidFactory as SymfonyUuidFactory;
+
+final class UuidFactory implements IdFactoryInterface
+{
+    public function __construct(private SymfonyUuidFactory $uuidFactory)
+    {
+        // The body is not required
+    }
+
+    public function create(): int|string
```
alexndlm, line 17: When can it be `int`?

alexndlm (changes requested):

packages/EasyActivity/src/Bridge/Symfony/Uid/UuidFactory.php
mvlad: Fixed.

voodooism (approved):

akl-95 (approved):

roman-eonx (approved):

natepage (approved):

natepage (approved):

PR #938, opened 2022-05-19
[EasyWebhook] feature/easy-webhooks-add-config-for-timestamps-timezone

evannamon:
- Added timezone on DoctrineDbalStore DoctrineDbalResult as optional parameter for timestamps.
---
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | #...

### DIFF
--- a/packages/EasyWebhook/src/Stores/DoctrineDbalResultStore.php
+++ b/packages/EasyWebhook/src/Stores/DoctrineDbalResultStore.php
@@ -17,22 +17,27 @@ final class DoctrineDbalResultStore extends AbstractDoctrineDbalStore implements
     /**
      * @var string|null
      */
-    private $timestampTimezone;
+    private $timestamp;
 
     public function __construct(
         RandomGeneratorInterface $random,
         Connection $conn,
         DataCleanerInterface $dataCleaner,
         ?string $table = null,
-        ?string $timestampTimezone = null
+        ?string $timestamp = null
     ) {
-        $this->timestampTimezone = $timestampTimezone;
+        $this->timestamp = $timestamp;
         parent::__construct($random, $conn, $dataCleaner, $table ?? 'easy_webhook_results');
+    }
+
+    public function getTimestamp(): ?string
+    {
+        return $this->timestamp;
     }
 
     public function store(WebhookResultInterface $result): WebhookResultInterface
     {
-        $timezone = $this->timestampTimezone ?? 'UTC';
+        $timezone = $this->timestamp ?? 'UTC';
 
         $now = Carbon::now($timezone);
         $data = $this->getData($result, $now);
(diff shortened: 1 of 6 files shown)


packages/EasyWebhook/src/Stores/DoctrineDbalResultStore.php
```diff
@@ -14,18 +14,27 @@
...
+        ?string $timestampTimezone = null
     ) {
+        $this->timestampTimezone = $timestampTimezone;
         parent::__construct($random, $conn, $dataCleaner, $table ?? 'easy_webhook_results');
     }
 
     public function store(WebhookResultInterface $result): WebhookResultInterface
     {
-        $now = Carbon::now('UTC');
+        $timezone = $this->timestampTimezone ?? 'UTC';
```
josemujicap, line 35: Let's rename timestampTimezone to "timezone". Let's be specific. timestamp refers to a different concept.

Also let's add a getter method

natepage: @evannamon agree with @josemujicap please let's rename it to simply `timezone`

natepage (changes requested):

packages/EasyWebhook/src/Stores/DoctrineDbalResultStore.php
evannamon: @josemujicap & @natepage fixed variable, added getter and assertion for the added method.

packages/EasyWebhook/src/Stores/DoctrineDbalResultStore.php
```diff
@@ -14,18 +14,35 @@
 
 final class DoctrineDbalResultStore extends AbstractDoctrineDbalStore implements ResultStoreInterface
 {
+    /**
+     * @var string
+     */
+    private const DEFAULT_TIMEZONE = 'UTC';
```
natepage, line 20: @evannamon Instead of duplicating this constant, I think we should put it on the `StoreInterface`

natepage (changes requested):

packages/EasyWebhook/src/Stores/DoctrineDbalResultStore.php
evannamon: Yep, moved this constant to `StoreInterface`

packages/EasyWebhook/src/Stores/DoctrineDbalStore.php
```diff
@@ -82,9 +89,14 @@ public function generateWebhookId(): string
         return $this->random->uuidV4();
     }
 
+    public function getTimezone(): ?string
```
natepage, line 92: @evannamon is this method only for the tests? If yes, then please remove it as any public method becomes something we have to maintain as people can use it.
Your main test `testStoreWithTimezone` is IMO already doing everything we need to do.

natepage (changes requested):

natepage (approved):

PR #950, opened 2022-06-06
[EasyErrorHandler] Translate exception message

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Translate exception message before it will be sent to Bagsnag, displayed in logs or error 500 page.

### DIFF
--- a/packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
+++ b/packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
@@ -11,8 +11,10 @@ use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 
 final class ExceptionEventListener
 {
-    public function __construct(private ErrorHandlerInterface $errorHandler, private TranslatorInterface $translator)
-    {
+    public function __construct(
+        private ErrorHandlerInterface $errorHandler,
+        private TranslatorInterface $translator
+    ) {
     }
 
     public function __invoke(ExceptionEvent $event): void
--- a/packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
+++ b/packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
@@ -5,29 +5,31 @@ declare(strict_types=1);
 namespace EonX\EasyErrorHandler\Bridge\Symfony\Listener;
 
 use EonX\EasyErrorHandler\Interfaces\ErrorHandlerInterface;
+use EonX\EasyErrorHandler\Interfaces\Exceptions\TranslatableExceptionInterface;
+use EonX\EasyErrorHandler\Interfaces\TranslatorInterface;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 
 final class ExceptionEventListener
 {
-    /**
-     * @var \EonX\EasyErrorHandler\Interfaces\ErrorHandlerInterface
-     */
-    private $errorHandler;
-
-    public function __construct(ErrorHandlerInterface $errorHandler)
+    public function __construct(private ErrorHandlerInterface $errorHandler, private TranslatorInterface $translator)
     {
-        $this->errorHandler = $errorHandler;
     }
 
     public function __invoke(ExceptionEvent $event): void
     {
-        $this->errorHandler->report($event->getThrowable());
+        $throwable = $event->getThrowable();
+
+        if ($throwable instanceof TranslatableExceptionInterface) {
+            $throwable->setMessage($this->translator->trans($throwable->getMessage(), $throwable->getMessageParams()));
+        }
+
+        $this->errorHandler->report($throwable);
 
         // Skip if format is html
         if ($event->getRequest()->getRequestFormat('') === 'html') {
             return;
         }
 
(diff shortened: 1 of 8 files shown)


packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
```diff
@@ -5,29 +5,33 @@
...
+    ) {
     }
 
     public function __invoke(ExceptionEvent $event): void
     {
-        $this->errorHandler->report($event->getThrowable());
+        $throwable = $event->getThrowable();
+
+        if ($throwable instanceof TranslatableExceptionInterface) {
+            $throwable->setMessage($this->translator->trans($throwable->getMessage(), $throwable->getMessageParams()));
```
roman-eonx, line 25: I think we shouldn't translate the message BEFORE passing a throwable to the error handler. Because in this case, we will try to translate it again in `ExtendedExceptionBuilder`. At least, when we call the `render` method.

If we want the message to be translated when we call the `report` method, then we should do this on lower levels. Also, if we translate only in the Symfony Bridge, your fix won't work for Laravel applications.

@natepage could you please advise a better place to do translation, common for usual logs and Bugsnags? I would say it should be `ErrorDetailsResolver` (in both, `resolveSimpleDetails` and `resolveExtendedDetails` methods, to translate both, `message` and `user_message`). But `LoggerReporter` logs the message "as is" (without passing it through the resolver).

roman-eonx (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
```diff
@@ -5,29 +5,33 @@
...
-     * @var \EonX\EasyErrorHandler\Interfaces\ErrorHandlerInterface
-     */
-    private $errorHandler;
-
-    public function __construct(ErrorHandlerInterface $errorHandler)
-    {
-        $this->errorHandler = $errorHandler;
+    public function __construct(
+        private ErrorHandlerInterface $errorHandler,
+        private TranslatorInterface $translator
```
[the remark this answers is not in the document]
alexndlm: Here I want `EonX\EasyErrorHandler\Interfaces\TranslatorInterface`.

`EonX\EasyCore\Bridge\Symfony\Traits\TranslatorAwareTrait` works with `Symfony\Contracts\Translation\TranslatorInterface`.

packages/EasyErrorHandler/src/Bridge/Symfony/Listener/ExceptionEventListener.php
alexndlm: We can remove transaction in `ExtendedExceptionBuilder`.

natepage: @alexndlm @roman-eonx thank you for this PR. The real question is where exactly do we want to see this translated message.

- in logs: is it ok to have it in the context?
- in bugsnag: is it ok to have it in the error tab?
- error 500 page: what is that?

roman-eonx: @natepage well, I'd say we should see them as "main" messages in both, logs and Bugsnags. But I see that it would be probably harder to implement than with the context or the error tab.

alexndlm: I have removed the translation of the exception message from `ExtendedExceptionBuilder`.

roman-eonx: I still don't like the current solution because:
1. It will work only for Symfony.
2. It introduces a BC because of the changed `TranslatableExceptionInterface` that can be potentially implemented in some of our projects.

@natepage do you have any ideas on where it would be better to translate the message if we want it to work the way described above?

natepage: @roman-eonx I'm not sure exactly where, but definitely not in the symfony event listener, it has to be part of the core logic of the package and not in a particular bridge. I need to think about it but I've been really busy with AVC lately sorry. I promise I will come back to this shortly.

natepage: @alexndlm @roman-eonx one more question, what message are you expecting to see there?

- `message`: internal message for devs
- `userMessage`: external message for end users

roman-eonx: @natepage the first one. The second one is not that important for debugging and can be left untranslated.

natepage: @alexndlm @roman-eonx I've had a look at this and I have an idea, I will create a separate PR for you to review 😉

alexndlm:
We decided to use another approach.

Please see https://github.com/eonx-com/easy-monorepo/pull/959

PR #963, opened 2022-06-14
[EasyBatch] Add kernel.reset tag to BatchProcessor

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyBatch/src/BatchObjectManager.php
+++ b/packages/EasyBatch/src/BatchObjectManager.php
@@ -18,7 +18,6 @@ use EonX\EasyBatch\Iterator\BatchItemIterator;
 use EonX\EasyBatch\Iterator\IteratorConfig;
 use EonX\EasyBatch\Persisters\BatchPersister;
 use EonX\EasyBatch\Processors\BatchProcessor;
-use EonX\EasyEventDispatcher\Interfaces\EventDispatcherInterface;
 
 final class BatchObjectManager implements BatchObjectManagerInterface
 {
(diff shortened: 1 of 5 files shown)


packages/EasyBatch/src/BatchObjectManager.php
```diff
@@ -73,10 +72,6 @@ public function approve(BatchObjectInterface $batchObject): BatchObjectInterface
                 $this->approve($this->batchItemRepository->findOrFail($batchObject->getParentBatchItemId()));
             }
 
-            // Approving a batch is possible only when all its batchItems succeeded and approval was required
```
alexndlm, line 76: We have dispatch event next in a `\EonX\EasyBatch\Processors\BatchProcessor::dispatchBatchEvent`

natepage: @alexndlm this event dispatch is to handle the case when we are approving the batch itself which in this case we don't go through the batch processor, or maybe I don't understand the issue.
Could you please explain how to reproduce the bug? Thanks

natepage (changes requested):

packages/EasyBatch/src/BatchObjectManager.php
alexndlm: I have rollbacked this. I will research this issue in addition.

akl-95 (approved):

akl-95 (approved):

natepage:
@alexndlm I'm not convinced with this solution. It feels really obscure like this, I think it should be more explicit that we need to reset services before processing a batchItem and therefore I think it should be in the Messenger middleware.

alexndlm:
@natepage I have refactored. Please check.

natepage:
@alexndlm I prefer the solution from #969 
what do you think?

alexndlm:
> @alexndlm I prefer the solution from #969
> what do you think?

@natepage This will not now work.
We need to clear the cache after the item is processing.

After $this->processBatchItemWithLock (line 57) we will have items in cache.

natepage:
Thanks @alexndlm for this.
Was done in #969

PR #969, opened 2022-06-21
[EasyBatch] Clear batch processor cache before processing item

natepage:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes


packages/EasyBatch/src/Bridge/Symfony/Messenger/ProcessBatchItemMiddleware.php
```diff
@@ -45,6 +47,9 @@ public function handle(Envelope $envelope, StackInterface $stack): Envelope
             return $func();
         }
 
+        // Make sure to have a clean batch processor before processing item to prevent caching issues.
+        $this->batchProcessor->reset();
```
alexndlm, line 51: This will not now work.
We need to clear the cache after the item is processing.

After `$this->processBatchItemWithLock` (line 57) we will have items in cache.

alexndlm (changes requested):

PR #982, opened 2022-06-29
[EasyErrorHandler] feature/error-handler-bugsnag-ignored-exceptions-resolver

alexandrtspl:
* ignoredExceptionsResolver

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    

Added ~~bugsnag_ignored_exceptions_resolver config option~~ service `BugsnagIgnoreExceptionResolver` to decide to ignore exception or not.

~~And ValidationErrorResponseBuilder moved due to task: https://eonx.atlassian.net/browse/PLU-4584~~


roman-eonx:
@natepage in this PR we implement two features:
1. In `ValidationErrorResponseBuilder` we convert every exception related to property types (e.g. a wrong data type/format passed in an API request) into a validation error response. This is a workaround for the biggest and most annoying API Platform issue: when 500 is returned instead of 400 for an API request containing incorrect data in the payload. I'm pretty sure this should be the default behavior for all our projects (until API Platform solves their issue), so we definitely need to enable it by default.
2. With `BugsnagIgnoredExceptionsResolver` we ignore all such errors in `BugsnagReporter`, so we don't send useless errors to BS. And here I want your confirmation about enabling this resolver by default. Is there any chance that some project still wants to see `This value is not a valid date/time.`, `Invalid IRI "..."`, `The type of the "..." attribute must be "...", "..." given`, or similar errors in BS?

packages/EasyErrorHandler/src/Bridge/Bugsnag/BugsnagReporter.php
```diff
@@ -13,35 +14,33 @@
...
-     * @var \Bugsnag\Client
-     */
-    private $bugsnag;
+    private Client $bugsnag;
 
     /**
      * @var string[]
      */
-    private $ignoreExceptions;
+    private ?array $ignoreExceptions;
```
natepage, line 22: This cannot be null as we default it in the constructor.

packages/EasyErrorHandler/src/Bridge/Bugsnag/BugsnagReporter.php
```diff
@@ -52,12 +51,20 @@ public function __construct(
...
+            foreach ($this->ignoreExceptions as $ignoreClass) {
+                if (\is_a($exceptionClass, $ignoreClass, true)) {
+                    return;
+                }
             }
         }
 
+        if ($this->ignoreExceptionsResolver !== null
+            && \is_a($this->ignoreExceptionsResolver, BugsnagIgnoreExceptionsResolverInterface::class, true) === true
+            && $this->ignoreExceptionsResolver::shouldIgnore($throwable) === true) {
```
natepage, line 64: I like the idea of delegating the decision to an object as it will allow applications to extend this logic to fit their needs. However, I don't like the current implementation, defining a config pointing to a class with a static method seems complicated and not really straight forward, and on the top of that it restricts the possibilities of the applications how to implement their logic.

It would be much better to use the "service approach" for this:

- create a dedicated service to decide if an exception should be ignored in bugsnag
- inject this service in this BugsnagReporter
- create a default implementation for it (could return false by default)

It will simplify the logic in the reporter as it will always expect to have an instance of this service to make the decision.
It will simplify the process for applications to extend/replace the default logic by decorating/replacing the service, AND will allow them to use all the dependency injection features to implement their own logic!

- no need for new config and param
- simpler logic in bugsnag reporter
- unlimited possibilites for bugsnag ignore logic

packages/EasyErrorHandler/src/Bridge/Bugsnag/BugsnagReporterProvider.php
```diff
@@ -54,7 +31,8 @@ public function getReporters(): iterable
             $this->bugsnag,
             $this->errorLogLevelResolver,
             $this->threshold,
-            $this->ignoredExceptions
+            $this->ignoredExceptions,
+            $this->ignoredExceptionsResolver,
```
natepage, line 35: We are not using trailing comas on function arguments.

packages/EasyErrorHandler/src/Builders/ValidationErrorResponseBuilder.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Builders;
+
+use ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
```
natepage, line 8: We can't do this or it means this package won't work on application that don't have api-platform, need to revisit the solution.

natepage (changes requested):

packages/EasyErrorHandler/src/Bridge/Bugsnag/BugsnagReporterProvider.php
alexandrtspl: Fixed

packages/EasyErrorHandler/src/Bridge/Bugsnag/BugsnagReporter.php
alexandrtspl: Fixed

packages/EasyErrorHandler/src/Bridge/Bugsnag/BugsnagReporter.php
alexandrtspl: Refactored a little. Please check it

packages/EasyErrorHandler/src/Builders/ValidationErrorResponseBuilder.php
alexandrtspl: Removed

roman-eonx: @alexandrtspl we can add `EonX\EasyErrorHandler\Bridge\ApiPlatform\Builder\ValidationErrorResponseBuilder` that will be specific to API Platform.

alexandrtspl: Code added to Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php

roman-eonx: @natepage as soon as `Bridge/Symfony/Builder/ApiPlatformValidationExceptionBuilder.php` is specific to API Platform and is enabled only when the specific class exists, I think it should be fine to go this way.

roman-eonx (dismissed):

packages/EasyErrorHandler/src/Builders/ValidationErrorResponseBuilder.php
natepage: Yes I think so, thanks

natepage:
@alexandrtspl @roman-eonx please fix the conflicts and we are good to go

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformErrorResponseBuilder.php
```diff
@@ -0,0 +1,187 @@
...
+use ApiPlatform\Core\Bridge\Symfony\Validator\Exception\ValidationException;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use EonX\EasyErrorHandler\Builders\AbstractErrorResponseBuilder;
+use EonX\EasyErrorHandler\Interfaces\TranslatorInterface;
+use Symfony\Component\Serializer\Exception\MissingConstructorArgumentsException;
+use Symfony\Component\Serializer\Exception\UnexpectedValueException;
+use Symfony\Component\Validator\Constraints\NotNull;
+use Throwable;
+
+final class ApiPlatformErrorResponseBuilder extends AbstractErrorResponseBuilder
```
akl-95, line 16: I believe we shouldn't collapse two builders into one. I'd suggest the following classnames and its responsibilities:

 - `ApiPlatformValidationExceptionResponseBuilder` is responsible for building correct response for `ValidationException`.
 - `ApiPlatformValidationErrorResponseBuilder` is responsible for building correct response for any error that is treated as validation exception.

Extra points for this idea:
 - `ApiPlatformValidationErrorResponseBuilder::supports` makes no sense for `ValidationException` because it's already ignored by `bugsnagIgnoredExceptions` configuration
 - `ApiPlatformValidationExceptionResponseBuilder` doesn't need any `MESSAGE_PATTERN_...` consts for its logic

Splitting of these responsibilities would make code cleaner I believe.

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -56,6 +58,9 @@
             'template' => '@EasyErrorHandlerSymfony/Collector/error_handler_collector.html.twig',
         ]);
 
+    // BugsnagIgnoreExceptionsResolver
+    $services->set(BugsnagIgnoreExceptionsResolverInterface::class, BugsnagIgnoreExceptionsResolver::class);
```
akl-95, line 62: What about configuration, similar to `bugsnagIgnoredExceptions`? So we can see `bugsnagIgnoredExceptions` and `bugsnagIgnoreExceptionsResolver` options in one place and explicitly see all the ignore-related configuration?

akl-95 (changes requested):

itorgov (approved):

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformErrorResponseBuilder.php
alexandrtspl: Done. Please check

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationErrorResponseBuilder.php
```diff
@@ -0,0 +1,125 @@
...
+        return match ($throwable::class) {
+            InvalidArgumentException::class => \preg_match(self::MESSAGE_PATTERN_TYPE_ERROR, $message) === 1,
+            MissingConstructorArgumentsException::class =>
+                \preg_match(self::MESSAGE_PATTERN_NO_PARAMETER, $message) === 1,
+            UnexpectedValueException::class =>
+                \preg_match(self::MESSAGE_PATTERN_INVALID_DATE, $message) === 1 ||
+                \preg_match(self::MESSAGE_PATTERN_INVALID_IRI, $message) === 1 ||
+                \preg_match(self::MESSAGE_PATTERN_NESTED_DOCUMENTS_NOT_ALLOWED, $message) === 1 ||
+                \preg_match(self::MESSAGE_PATTERN_NOT_IRI, $message) === 1,
+            ValidationException::class => true,
```
akl-95, line 45: Do we really need to support `ValidationException` here? It seems we have all the required logic in `ApiPlatformValidationExceptionResponseBuilder`.

akl-95 (changes requested):

natepage:
@akl-95 @roman-eonx regarding ignoring all API platform exceptions by default, I think it's ok but it should be subject to a config or something so if someone doesn't want this they can easily disable it

packages/EasyErrorHandler/src/Bridge/Bugsnag/Resolvers/DefaultBugsnagIgnoreExceptionsResolver.php
```diff
@@ -0,0 +1,16 @@
...
+namespace EonX\EasyErrorHandler\Bridge\Bugsnag\Resolvers;
+
+use EonX\EasyErrorHandler\Bridge\Bugsnag\Interfaces\BugsnagIgnoreExceptionsResolverInterface;
+use Throwable;
+
+final class DefaultBugsnagIgnoreExceptionsResolver implements BugsnagIgnoreExceptionsResolverInterface
+{
+    public function shouldIgnore(Throwable $throwable): bool
+    {
+        return false;
```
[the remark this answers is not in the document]
alexandrtspl: Done. Please check it

packages/EasyErrorHandler/tests/Bridge/Bugsnag/BugsnagReporterTest.php
```diff
@@ -4,7 +4,9 @@
 
 namespace EonX\EasyErrorHandler\Tests\Bridge\Bugsnag;
```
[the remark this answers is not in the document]
alexandrtspl: Fixed

packages/EasyErrorHandler/src/Bridge/Symfony/Builder/ApiPlatformValidationErrorResponseBuilder.php
alexandrtspl: No, removed

akl-95 (approved):

natepage (approved):

PR #983, opened 2022-06-30
[EasyDoctrine] Fix timezone issue in CarbonImmutableDateTime

mvlad:
Currently out DBAL Type for `DateTime`'s doesn't take into account `TimeZone`'s.

As a result, `2022-01-01T10:00:00+09:00` will be processed in DB as `2022-01-01T10:00:00+00:00`. It affects `WHERE` conditions and stored values. 

New implementation is based on advice from [doctrine documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/cookbook/working-with-datetime.html#:~:text=By%20default%20Doctrine%20assumes%20that,or%20by%20calling%20date_default_timezone_set()%20.). 
Changes may break BC in theory.  But in practice everything should be fine because we use UTC in our projects.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no 
| Tests pass?   | yes

### DIFF
--- a/packages/EasyDoctrine/src/DBAL/Types/CarbonImmutableDateTimeMicrosecondsType.php
+++ b/packages/EasyDoctrine/src/DBAL/Types/CarbonImmutableDateTimeMicrosecondsType.php
@@ -5,14 +5,18 @@ declare(strict_types=1);
 namespace EonX\EasyDoctrine\DBAL\Types;
 
 use Carbon\CarbonImmutable;
+use DateTime;
 use DateTimeImmutable;
 use DateTimeInterface;
+use DateTimeZone;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\ConversionException;
 use Doctrine\DBAL\Types\DateTimeImmutableType;
 
 final class CarbonImmutableDateTimeMicrosecondsType extends DateTimeImmutableType
 {
+    private static ?DateTimeZone $utc = null;
+
     /**
      * @var string
      */
@@ -44,8 +48,8 @@ final class CarbonImmutableDateTimeMicrosecondsType extends DateTimeImmutableTyp
             return $value;
         }
 
-        if ($value instanceof DateTimeInterface) {
-            return $value->format(self::FORMAT_PHP_DATETIME);
+        if ($value instanceof DateTimeImmutable || $value instanceof DateTime) {
+            return $value->setTimezone(self::getUtc())->format(self::FORMAT_PHP_DATETIME);
         }
 
         throw ConversionException::conversionFailedInvalidType($value, $this->getName(), ['null', 'DateTimeInterface']);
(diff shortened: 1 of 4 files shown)


packages/EasyDoctrine/src/DBAL/Types/CarbonImmutableDateTimeMicrosecondsType.php
```diff
@@ -5,14 +5,18 @@
...
 use Carbon\CarbonImmutable;
+use DateTime;
 use DateTimeImmutable;
 use DateTimeInterface;
+use DateTimeZone;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\ConversionException;
 use Doctrine\DBAL\Types\DateTimeImmutableType;
 
 final class CarbonImmutableDateTimeMicrosecondsType extends DateTimeImmutableType
```
roman-eonx, line 16: How about applying a similar fix to `DateTimeImmutableMicrosecondsType`?

roman-eonx (changes requested):

egor-dev (dismissed):

packages/EasyDoctrine/src/DBAL/Types/CarbonImmutableDateTimeMicrosecondsType.php
mvlad: Done.

packages/EasyDoctrine/src/DBAL/Types/CarbonImmutableDateTimeMicrosecondsType.php
```diff
@@ -5,14 +5,18 @@
...
 {
+    private static ?DateTimeZone $utc = null;
```
itorgov, line 18: Do we really need to make it static?

itorgov (changes requested):

egor-dev (approved):

packages/EasyDoctrine/src/DBAL/Types/CarbonImmutableDateTimeMicrosecondsType.php
itorgov: Yes, but is it really that helpful? Why don't we use `static` properties everywhere? :)

itorgov: We could just initialise it in the constructor, isn't it?

itorgov: More than that we have started to use Swoole and keeping the state is not a good idea now.

roman-eonx: I will let @mvlad reply to this. But as of Swoole, it doesn't matter here, as this "state" is kinda constant and useful.

mvlad: I copied this static method from the [Doctrine documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/cookbook/working-with-datetime.html#:~:text=By%20default%20Doctrine%20assumes%20that,or%20by%20calling%20date_default_timezone_set()%20.). It's needed to optimize the logic.

> We could just initialise it in the constructor, isn't it?

No, `Doctrine\DBAL\Types\Type::__construct` [is final](https://github.com/doctrine/dbal/blob/3.3.x/src/Types/Type.php#L56).

> More than that we have started to use Swoole and keeping the state is not a good idea now.

`DateTimeZone` here is kinda constant as Roman said. This value can't be changed, also `getUtc` is private and can't be used outside the `CarbonImmutableDateTime`. So I don't see even theoretical issues with this logic.

natepage (approved):

PR #999, opened 2022-07-15
[EasySecurity] Add support for role based access

mvlad:
Added the `role()` function similar to `permission()`. Now we can check security using a function like `is_granted(role('ROLE_CUSTOMER'))`.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes


packages/EasySecurity/src/Bridge/Symfony/Security/RoleExpressionFunctionProvider.php
```diff
@@ -0,0 +1,61 @@
...
+use Psr\Log\NullLogger;
+use Symfony\Component\ExpressionLanguage\ExpressionFunction;
+use Symfony\Component\ExpressionLanguage\ExpressionFunctionProviderInterface;
+
+final class RoleExpressionFunctionProvider implements ExpressionFunctionProviderInterface
+{
+    /**
+     * @var mixed[]
+     */
+    private array $cached = [];
```
egor-dev, line 18: I don't see who sets that. Do we need it?

egor-dev (changes requested):

roman-eonx (approved):

egor-dev (approved):

packages/EasySecurity/src/Bridge/Symfony/Security/RoleExpressionFunctionProvider.php
itorgov: See below on the line 48 (`$this->cached[$role] = \constant($constant);`).

natepage (approved):

PR #1003, opened 2022-07-21
[EasySecurity] Add deprecation for SecurityContextInterface autowiring usage

BOB41K1987:
- Add deprecation for SecurityContextInterface autowiring usage

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes

### DIFF
--- a/packages/EasySecurity/src/Bridge/Symfony/DependencyInjection/Compiler/RegisterSecurityContextPass.php
+++ b/packages/EasySecurity/src/Bridge/Symfony/DependencyInjection/Compiler/RegisterSecurityContextPass.php
@@ -39,7 +39,8 @@ final class RegisterSecurityContextPass implements CompilerPassInterface
             ->setDeprecated(
                 'eonx-com/easy-security',
                 '4.1.37',
-                'The "%service_id%" service autowiring is deprecated and will be removed in 5.0. Use SecurityContextResolverInterface::resolveContext instead.'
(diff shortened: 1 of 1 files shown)


natepage:
@BOB41K1987 thanks for that, could we please also add a CompilerPass to the Symfony bridge to detect any service having a dependency directly on the security context and trigger a deprecation for those? Thanks

roman-eonx:
> @BOB41K1987 thanks for that, could we please also add a CompilerPass to the Symfony bridge to detect any service having a dependency directly on the security context and trigger a deprecation for those? Thanks

@natepage actually, this was just a PoC. And we've found an issue with this approach: when we get the SecurityContext service directly from the container in tests (to arrange some test data), we trigger this deprecation notice. So, probably we will revise the current solution.

roman-eonx:
> > @BOB41K1987 thanks for that, could we please also add a CompilerPass to the Symfony bridge to detect any service having a dependency directly on the security context and trigger a deprecation for those? Thanks
> 
> @natepage actually, this was just a PoC. And we've found an issue with this approach: when we get the SecurityContext service directly from the container in tests (to arrange some test data), we trigger this deprecation notice. So, probably we will revise the current solution.

@natepage alright, we've found a way to work around this. Everything is ok now, we can merge this PR.

PR #1013, opened 2022-07-29
[EasyTest] Add MessengerAssertionsTrait

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Add MessengerAssertionsTrait

### DIFF
--- a/packages/EasyTest/src/Trait/Messenger/MessengerAssertionsTrait.php
+++ b/packages/EasyTest/src/Trait/Messenger/MessengerAssertionsTrait.php
@@ -158,7 +158,7 @@ trait MessengerAssertionsTrait
     public static function getMessagesSentToTransport(string $transportName, ?string $messageClass = null): array
     {
         /** @var \Symfony\Component\Messenger\Transport\InMemoryTransport $transport */
-        $transport = KernelTestCase::getContainer()->get("messenger.transport.$transportName");
+        $transport = KernelTestCase::getContainer()->get("messenger.transport.${transportName}");
 
         $messages = [];
         foreach ($transport->getSent() as $envelope) {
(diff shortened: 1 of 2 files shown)


packages/EasyTest/src/Trait/Messenger/MessengerAssertionsTrait.php
```diff
@@ -0,0 +1,211 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyTest\Trait\Messenger;
```
akl-95, line 5: Please fix to `Traits` to be consistent.

akl-95, line 5: Do you think we need `Messenger` level right now? It seems excessive to me for now.

akl-95 (changes requested):

packages/EasyTest/src/Trait/Messenger/MessengerAssertionsTrait.php
alexndlm: Fixed

alexndlm: Exactly, I agree. Refactored.

akl-95 (approved):

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -0,0 +1,211 @@
...
+                return true;
+            }
+        );
+
+        Assert::assertCount($messagesCount ?? 1, $envelopes);
+    }
+
+    public static function consumeAsyncMessages(): void
+    {
+        self::runMessengerWorker('async', 'default');
```
roman-eonx, line 119: Let's extract all the magic strings into constants.

roman-eonx (changes requested):

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: For this, we need to add an interface or some abstract class for tests.
I think it will be overcomplicated for this case.

WDYT?

roman-eonx: No need for an interface, just add constants to this trait. Simply to not repeat these values through the trait code.

alexndlm: I meant that this is a trait.

We can add static properties, but this also does not look good.

roman-eonx: Indeed, I forgot about the inability to add constants to a trait. Ok, let's keep it as is.

roman-eonx (approved):

natepage (approved):

PR #1023, opened 2022-08-17
[EasyUtils] Added credit card number validation

zeromodule:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    

Added credit card number validation to `CreditCardNumberStringSanitizer` to prevent from sanitizing strings that look like credit card number, but they are not really.

The logic (and card types support) of `CreditCardNumberValidator` is the same as https://symfony.com/doc/current/reference/constraints/CardScheme.html


packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyUtils\CreditCard;
+
+final class CreditCardNumberValidator
```
akl-95, line 7: Why did you decide to not use the external package?

zeromodule: Because code of this library was too bad and incompatible with PHP 8. 
Considering this and that code size is not very big, I decided to clean it, adopt it and not to require external library.

packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
```diff
@@ -0,0 +1,159 @@
...
+            'pattern' => '/^35/',
+            'length' => [16],
+            'luhn' => true,
+        ],
+    ];
+
+    public function isCreditCardNumberValid(string $number): bool
+    {
+        // Strip non-numeric characters
+        $number = preg_replace('/\D/', '', $number);
```
alexandrtspl, line 84: If validator removes non-numeric character inside $number, card number like: '5019 5555**DD** 4444 5555' will be correct. Is it ok?

zeromodule: @alexandrtspl good point. I don't think it's ok. 
I guess we should strip spaces, dashes, dots and slashes. WDYT?

alexandrtspl: I agree. By may be spaces and dashes will be anough. But it's not important (dots and slashes).

packages/EasyUtils/src/SensitiveData/StringSanitizers/CreditCardNumberStringSanitizer.php
```diff
@@ -17,8 +19,14 @@ public function sanitizeString(string $string, string $maskPattern, array $keysT
             return $string;
         }
 
+        $creditCardNumberValidator = new CreditCardNumberValidator();
```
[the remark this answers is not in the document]
zeromodule: We can, but it'll break the code in https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasyUtils/src/Bridge/Laravel/EasyUtilsServiceProvider.php#L79 because currently all sanitizers have same constructor. 

Thinking about the best way to solve this. WDYT?

packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
zeromodule: @alexandrtspl there is test case https://github.com/eonx-com/easy-monorepo/blob/master/packages/EasyUtils/tests/AbstractSensitiveDataSanitizerTestCase.php where dots and slashes are also checked, so I should say it's better to cut them out too.

natepage:
@akl-95 @alexandrtspl Do you guys need anything on this to get it done? Because I like the idea and would love to see it released

akl-95:
> @akl-95 @alexandrtspl Do you guys need anything on this to get it done? Because I like the idea and would love to see it released

@natepage free time only. We're going to finish it.

packages/EasyUtils/src/SensitiveData/StringSanitizers/CreditCardNumberStringSanitizer.php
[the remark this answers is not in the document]
zeromodule: done

packages/EasyUtils/src/Bridge/Laravel/EasyUtilsServiceProvider.php
```diff
@@ -22,12 +24,7 @@
 
 final class EasyUtilsServiceProvider extends ServiceProvider
 {
-    private const DEFAULT_STRING_SANITIZERS = [
-        AuthorizationStringSanitizer::class,
-        CreditCardNumberStringSanitizer::class,
-        JsonStringSanitizer::class,
-        UrlStringSanitizer::class,
-    ];
+    private const STRING_SANITIZER_DEFAULT_PRIORITY = 1000;
```
akl-95, line 27: There was `10000` value passed, please fix.

packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
```diff
@@ -0,0 +1,148 @@
...
```
akl-95, line 73: Please fix `preg_replace` -> `\preg_replace`. The same goes for other built-in functions usage in this PR.

packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
```diff
@@ -0,0 +1,148 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyUtils\CreditCard;
+
+final class CreditCardNumberValidator implements CreditCardNumberValidatorInterface
```
akl-95, line 7: Please add a link to the source of this code in the PR description. Just for other reviewers.

akl-95 (changes requested):

packages/EasyUtils/src/Bridge/Laravel/EasyUtilsServiceProvider.php
zeromodule: Fixed

packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
zeromodule: Fixed

packages/EasyUtils/src/CreditCard/CreditCardNumberValidator.php
zeromodule: Done

akl-95 (dismissed):

packages/EasyUtils/src/CreditCard/CreditCardNumberValidatorInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyUtils\CreditCard;
+
+interface CreditCardNumberValidatorInterface
```
alexndlm, line 7: @zeromodule, why we can't use https://symfony.com/doc/current/reference/constraints/CardScheme.html?

alexndlm (changes requested):

packages/EasyUtils/src/CreditCard/CreditCardNumberValidatorInterface.php
zeromodule: As I understand, we can't just inject Symfony's `ValidatorInterface` to `CreditCardNumberStringSanitizer` because the code should be cross-platform and work identically in Symfony and Laravel.

alexndlm (approved):

packages/EasyUtils/src/CreditCard/CreditCardNumberValidatorInterface.php
alexndlm: Got it, thanks.

It is a little dirty solution, but I can't imagine a better one. 😂

alexandrtspl (approved):

packages/EasyUtils/src/CreditCard/CreditCardNumberValidatorInterface.php
roman-eonx: @zeromodule just in case, please verify that your solution works completely identically to Symfony's `CardSchemeValidator`. This is important not to let some specific card numbers be passed to logs.

roman-eonx (dismissed):

akl-95 (approved):

alexndlm (approved):

roman-eonx (approved):

natepage (approved):

PR #1028, opened 2022-08-31
[EasyActivity] ActivityLogEntryMessage does not implement WithLockDataInterface

mvlad:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyActivity/src/Bridge/Symfony/Messenger/ActivityLogEntryMessage.php
+++ b/packages/EasyActivity/src/Bridge/Symfony/Messenger/ActivityLogEntryMessage.php
@@ -5,9 +5,16 @@ declare(strict_types=1);
 namespace EonX\EasyActivity\Bridge\Symfony\Messenger;
 
 use EonX\EasyActivity\ActivityLogEntry;
+use EonX\EasyLock\Interfaces\LockDataInterface;
+use EonX\EasyLock\Interfaces\WithLockDataInterface;
+use EonX\EasyLock\LockData;
 
-final class ActivityLogEntryMessage
+final class ActivityLogEntryMessage implements WithLockDataInterface
 {
+    private const LOCK_RESOURCE = 'activity_log_%s_%s';
+
+    private const LOCK_TTL_SEC = 3600.0;
+
     /**
      * @var \EonX\EasyActivity\ActivityLogEntry
      */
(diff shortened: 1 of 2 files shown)


roman-eonx (approved):

itorgov (approved):

packages/EasyActivity/src/Bridge/Symfony/Messenger/ActivityLogEntryMessage.php
```diff
@@ -5,9 +5,16 @@
 namespace EonX\EasyActivity\Bridge\Symfony\Messenger;
 
 use EonX\EasyActivity\ActivityLogEntry;
+use EonX\EasyLock\Interfaces\LockDataInterface;
+use EonX\EasyLock\Interfaces\WithLockDataInterface;
+use EonX\EasyLock\LockData;
 
-final class ActivityLogEntryMessage
+final class ActivityLogEntryMessage implements WithLockDataInterface
```
natepage, line 12: will that fail if EasyLock isn't installed?

natepage (changes requested):

packages/EasyActivity/src/Bridge/Symfony/Messenger/ActivityLogEntryMessage.php
mvlad: Yes, the following error occurs without EasyLock:
```
Error: Interface "EonX\EasyLock\Interfaces\WithLockDataInterface" not found
```
https://github.com/eonx-com/easy-monorepo/runs/8109257071?check_suite_focus=true#step:7:17.

roman-eonx: @natepage we suggest installing this package in `composer.json`, similar to `symfony/messenger`. A project can use the package without async logic or with its own implementation,

natepage: @roman-eonx so this message isn't triggered by default? do we have like a config option to enable async?

roman-eonx: @natepage `\EonX\EasyActivity\Bridge\Symfony\Messenger\AsyncDispatcher` is the default implementation of `\EonX\EasyActivity\Interfaces\AsyncDispatcherInterface`. And `\EonX\EasyActivity\Logger\AsyncActivityLogger` is the default implementation of `\EonX\EasyActivity\Interfaces\ActivityLoggerInterface`. Both are set in the Symfony bridge config. So, this is the default configuration for a Symfony project. But a project can implement its own dispatcher without Messenger or without locking. Or even use synchronous dispatching.

natepage: ok thank you for the explanation

natepage (approved):

PR #1034, opened 2022-09-07
[EasySwoole] `StreamedResponse` not streaming

mvlad:
EasySwoole surrounds `Application::handle` with output buffering to support `echo`, `var_dump`, etc.
As a result `StreamedResponse` doesn't work as expected. Instead of sending a content to the client, this content is buffered and `Allowed memory size of 268435456 bytes exhausted` occurs for big responses. 

This PR removes output buffering. `echo`/`var_dump` will be available in the docker container's stdout.
As we don't use `echo`/`var_dump` for normal responses, changes should not lead to BC breaks.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasySwoole/src/Helpers/HttpFoundationHelper.php
+++ b/packages/EasySwoole/src/Helpers/HttpFoundationHelper.php
@@ -52,8 +52,7 @@ final class HttpFoundationHelper
     public static function reflectHttpFoundationResponse(
         HttpFoundationResponse $hfResponse,
         Response $response,
-        int $chunkSize,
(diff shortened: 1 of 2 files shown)


roman-eonx (approved):

egor-dev (approved):

roman-eonx (approved):

mvlad:
The issue was in https://github.com/symfony/symfony/pull/45476. It's enough to upgrade Symfony.

PR #1035, opened 2022-09-07
[EasyCore] Added support for multiple classes to AbstractInputDataTransformer

zeromodule:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    

Added support for multiple classes to AbstractInputDataTransformer to cover some specific cases when we have to use one InputDataTransformer for several API resources with a common ancestor or for several DTO classes.

### DIFF
--- a/packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
+++ b/packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
@@ -66,8 +66,14 @@ abstract class AbstractInputDataTransformer implements DataTransformerInterface
      */
     abstract protected function doTransform(object $object, ?array $context = null): object;
 
+    /**
+     * @return string[]|string
+     */
     abstract protected function getApiResourceClass(): array|string;
 
+    /**
+     * @return string[]|string
+     */
     abstract protected function getInputClass(): array|string;
 
     protected function isValidationNeeded(): bool
--- a/packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
+++ b/packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
@@ -29,13 +29,23 @@ abstract class AbstractInputDataTransformer implements DataTransformerInterface
     public function supportsTransformation(mixed $data, string $to, ?array $context = null): bool
     {
         $apiResourceClass = $this->getApiResourceClass();
-
-        if ($data instanceof $apiResourceClass) {
-            return false;
+        if (is_string($apiResourceClass)) {
+            $apiResourceClass = [$apiResourceClass];
         }
 
-        return is_a($to, $apiResourceClass, true) &&
-            is_a($context['input']['class'] ?? null, $this->getInputClass(), true);
+        foreach ($apiResourceClass as $class) {
+            if ($data instanceof $class) {
+                return false;
+            }
+        }
+
+        $inputClass = $this->getInputClass();
+        if (is_string($inputClass)) {
+            $inputClass = [$inputClass];
+        }
+
+        return in_array($to, $apiResourceClass, true) &&
+            in_array($context['input']['class'] ?? null, $inputClass, true);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
```diff
@@ -34,7 +34,8 @@ public function supportsTransformation(mixed $data, string $to, ?array $context
             return false;
         }
 
-        return $to === $apiResourceClass && ($context['input']['class'] ?? null) === $this->getInputClass();
+        return is_a($to, $apiResourceClass, true) === true &&
```
alexndlm, line 37: Let's remove ` === true `

alexndlm (changes requested):

packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
zeromodule: fixed

alexndlm: Thanks

alexndlm (approved):

packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
```diff
@@ -29,12 +29,23 @@ public function setValidator(ValidatorInterface $validator): self
     public function supportsTransformation(mixed $data, string $to, ?array $context = null): bool
     {
         $apiResourceClass = $this->getApiResourceClass();
```
akl-95, line 31: Can we rename `getApiResourceClass` -> `getApiResourceClasses` and always iterate over array?

packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
```diff
@@ -29,12 +29,23 @@ public function setValidator(ValidatorInterface $validator): self
...
+
+        foreach ($apiResourceClass as $class) {
+            if ($data instanceof $class) {
+                return false;
+            }
+        }
 
-        if ($data instanceof $apiResourceClass) {
-            return false;
+        $inputClass = $this->getInputClass();
```
akl-95, line 42: The same goes here. Can we make `getInputClass` -> `getInputClasses`?

akl-95 (changes requested):

packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
zeromodule: In this way we can break BC with the other projects code which is using old method name.

zeromodule: And for the same reason we can't always iterate over array because some code can currently return string in the implementation.

akl-95: Yeah, I see. The we shouldn't proceed with this solution, it has confusing contract.

ERuban: My opinion that nothing bad in such small BC break here, so we can rename and expect that this method always returns an array of classes.
The main thing is not to forget to add a BC break description to PR then.

packages/EasyCore/src/Bridge/Symfony/ApiPlatform/DataTransformer/AbstractInputDataTransformer.php
```diff
@@ -29,12 +29,23 @@ public function setValidator(ValidatorInterface $validator): self
...
+        if (is_string($inputClass)) {
+            $inputClass = [$inputClass];
         }
 
-        return $to === $apiResourceClass && ($context['input']['class'] ?? null) === $this->getInputClass();
+        return in_array($to, $apiResourceClass, true) &&
+            in_array($context['input']['class'] ?? null, $inputClass, true);
```
ERuban, line 48: `$context['input']['class'] ?? ''` looks better, imo

ERuban (changes requested):

alexndlm:
DataTranformer concept was removed in the API Platform 3.

PR #1051, opened 2022-10-11
[EasyDoctrine] Fix typehint in EntityEventSubscriber

BOB41K1987:
Fix typehint in EntityEventSubscriber

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #...   

In case the deferred dispatcher entity has a `OneToMany` `\Doctrine\Common\Collections\Collection` field and `cascade: ['persist']`, the argument can be `\Doctrine\ORM\PersistentCollection`.

### DIFF
--- a/packages/EasyDoctrine/src/Subscribers/EntityEventSubscriber.php
+++ b/packages/EasyDoctrine/src/Subscribers/EntityEventSubscriber.php
@@ -8,6 +8,7 @@ use DateTimeInterface;
 use Doctrine\ORM\Event\OnFlushEventArgs;
 use Doctrine\ORM\Event\PostFlushEventArgs;
 use Doctrine\ORM\Events;
+use Doctrine\ORM\PersistentCollection;
 use EonX\EasyDoctrine\Dispatchers\DeferredEntityEventDispatcherInterface;
 use EonX\EasyDoctrine\Interfaces\EntityEventSubscriberInterface;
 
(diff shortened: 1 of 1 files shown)


egor-dev (approved):

roman-eonx (approved):

packages/EasyDoctrine/src/Subscribers/EntityEventSubscriber.php
```diff
@@ -117,7 +118,7 @@ private function filterEntities(array $entities): array
      */
     private function getClearedChangeSet(array $changeSet): array
     {
-        return \array_filter($changeSet, static function (array $changeSetItem) {
+        return \array_filter($changeSet, static function (array|PersistentCollection $changeSetItem) {
```
natepage, line 121: Ok we change the typehint but don't we need to change the logic of the function based on the type given?

natepage (changes requested):

packages/EasyDoctrine/src/Subscribers/EntityEventSubscriber.php
BOB41K1987: Nope, as this function is only needed to filter unchanged `DateTimeInterface`

natepage: ok then I trust you

natepage (approved):

PR #1057, opened 2022-10-25
Add shared PhpStorm settings

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

We already use this approach in other projects. All works perfectly.

I have copied these settings from Payments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # Composer
 composer.lock
 /vendor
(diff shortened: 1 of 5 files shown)


alexndlm:
It was decided not to do this.

PR #1061, opened 2022-10-27
[EasyApiToken] Fix broken test and add compatibility with auth0-PHP >= 8.3

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/packages/EasyApiToken/src/External/Auth0JwtDriver.php
+++ b/packages/EasyApiToken/src/External/Auth0JwtDriver.php
@@ -106,6 +106,8 @@ final class Auth0JwtDriver implements JwtDriverInterface
         $config = new Auth0V8SdkConfiguration([
             'domain' => $this->domain,
             'clientId' => 'client_id',
+            'strategy' => Auth0V8SdkConfiguration::STRATEGY_API,
+            'audience' => [$this->audienceForEncode],
         ]);
         $verifier = new Auth0V8Token($config, $token, Auth0V8Token::TYPE_TOKEN);
         $exceptions = [];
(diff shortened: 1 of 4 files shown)


packages/EasyApiToken/tests/AbstractAuth0JwtTokenTestCase.php
```diff
@@ -45,16 +45,14 @@ protected function createAuth0JwtDriver(
         ?string $audienceForEncode = null,
         ?array $allowedAlgos = null
     ): JwtDriverInterface {
-        $auth0Driver = new Auth0JwtDriver(
+        return new Auth0JwtDriver(
             $validAudiences ?? static::$validAudiences,
             $authorizedIss ?? static::$authorizedIss,
-            'localhost',
+            'example.com',
```
alexndlm, line 51: We need this after this excellent commit https://github.com/auth0/auth0-PHP/commit/e89127a97ad72715e1a55c57fc3455146e12d9a8

packages/EasyApiToken/src/External/Auth0JwtDriver.php
```diff
@@ -106,6 +106,8 @@ public function decode(string $token): array
         $config = new Auth0V8SdkConfiguration([
             'domain' => $this->domain,
             'clientId' => 'client_id',
+            'strategy' => Auth0V8SdkConfiguration::STRATEGY_API,
```
alexndlm, line 109: We need this after this commit https://github.com/auth0/auth0-PHP/commit/a1fd98f8569649717cc50e18fb92521091c1312e

natepage: @alexndlm I think that would break applications that update the monorepo but not auth0

itorgov: Agree, we have to ensure that everything will work with version 8.0.

itorgov (changes requested):

packages/EasyApiToken/src/External/Auth0JwtDriver.php
alexndlm: @natepage these settings already exists in 8.0. From 8.3 these are required. Current changes should not impact apps.

itorgov (approved):

akl-95 (approved):

natepage (approved):

PR #1072, opened 2022-11-11
[EasyWebhook]  Do not resend message on CannotRerunWebhookException

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

We have two workers, and both fetch the same message.
|Worker 1| Worker 2|
| - | -
|Lock webhook|
|Super fast process webhook|
|Relase lock|
||
||Lock webhook
|| Try process webhook
||Throw `CannotRerunWebhookException`
||Relase lock
||Symfony resend the message with `1000 ms` delay
||
||Lock webhook
|| Try process webhook
||Throw `CannotRerunWebhookException`
||Relase lock
||Symfony resend the message with `2000 ms` delay
||
||Lock webhook
|| Try process webhook
||Throw `CannotRerunWebhookException`
||Relase lock
||Symfony resend the message with `4000 ms` delay

We have `\Symfony\Component\Messenger\Retry\MultiplierRetryStrategy`. Default `$maxRetries` is 3.

This strategy is configured here `src/vendor/symfony/framework-bundle/Resources/config/messenger.php:153` and then here `src/vendor/symfony/framework-bundle/DependencyInjection/FrameworkExtension.php:2091`.

So in our code, the `$maxRetries` is 3.

`MultiplierRetryStrategy` calculates the following delays:
- 1000 ms
- 2000 ms
- 4000 ms

These delays correlate with message receive time. See screens.

Снимок экрана 2022-11-11 в 14 19 59
Снимок экрана 2022-11-11 в 14 24 57

### DIFF
--- a/packages/EasyWebhook/src/Exceptions/CannotRerunWebhookException.php
+++ b/packages/EasyWebhook/src/Exceptions/CannotRerunWebhookException.php
@@ -4,10 +4,7 @@ declare(strict_types=1);
 
 namespace EonX\EasyWebhook\Exceptions;
 
-use Symfony\Component\Messenger\Exception\UnrecoverableExceptionInterface;
-
-final class CannotRerunWebhookException extends AbstractDoNotHandleMeException implements
-    UnrecoverableExceptionInterface
+final class CannotRerunWebhookException extends AbstractDoNotHandleMeException
 {
     // No body needed.
 }
--- a/packages/EasyWebhook/src/Exceptions/CannotRerunWebhookException.php
+++ b/packages/EasyWebhook/src/Exceptions/CannotRerunWebhookException.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace EonX\EasyWebhook\Exceptions;
 
-final class CannotRerunWebhookException extends AbstractDoNotHandleMeException
+use Symfony\Component\Messenger\Exception\UnrecoverableExceptionInterface;
+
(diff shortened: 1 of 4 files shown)


akl-95 (dismissed):

packages/EasyWebhook/src/Exceptions/CannotRerunWebhookException.php
```diff
@@ -4,7 +4,10 @@
 
 namespace EonX\EasyWebhook\Exceptions;
 
-final class CannotRerunWebhookException extends AbstractDoNotHandleMeException
+use Symfony\Component\Messenger\Exception\UnrecoverableExceptionInterface;
+
+final class CannotRerunWebhookException extends AbstractDoNotHandleMeException implements
```
natepage, line 9: @alexndlm Thank you for this PR, I completely understand why, but I don't think that this is the right way to implement it as this solution couples the package exception with a Symfony one outside of the bridge namespace. We must find another way from within the bridge code to achieve it.

natepage (changes requested):

packages/EasyWebhook/src/Exceptions/CannotRerunWebhookException.php
alexndlm: @natepage have done.

natepage (approved):

PR #1073, opened 2022-11-11
[All packages] Add gitattributes files (PoC)

ERuban:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    
| Fixed tickets | -   

- Add gitattributes files.

It prevents to import excluded folders while `composer install` call in projects. ([ref](https://alexbilbie.com/2012/11/exclude-objects-with-gitattributes/))
This will reduce `vendor` folder size and autoload.

### DIFF
--- a/packages/EasyActivity/.gitattributes
+++ b/packages/EasyActivity/.gitattributes
@@ -1,0 +1,3 @@
+docs/ export-ignore
+tests/ export-ignore
+phpunit.xml export-ignore
--- a/packages/EasyActivity/.gitattributes
+++ b/packages/EasyActivity/.gitattributes
@@ -1,3 +1,6 @@
 docs/ export-ignore
 tests/ export-ignore
 phpunit.xml export-ignore
+readme.md export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
--- a/packages/EasyApiPlatform/.gitattributes
+++ b/packages/EasyApiPlatform/.gitattributes
@@ -1,0 +1,3 @@
+docs/ export-ignore
+tests/ export-ignore
+phpunit.xml export-ignore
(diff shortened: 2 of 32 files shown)


packages/EasyActivity/.gitattributes
```diff
@@ -0,0 +1,3 @@
+docs/ export-ignore
```
alexndlm, line 1: I believe we should leave `docs`, as these files are mentioned in the `readme.md`

Or add `readme.md` for ignoring.

Wdyt?

packages/EasyActivity/.gitattributes
```diff
@@ -0,0 +1,3 @@
...
+tests/ export-ignore
+phpunit.xml export-ignore
```
alexndlm, line 3: What about adding `.gitignore`?

alexndlm (changes requested):

packages/EasyActivity/.gitattributes
ERuban: Agree. Ignored `readme.md`

packages/EasyActivity/.gitattributes
ERuban: Yeap, also have added `.gitattributes` file

packages/EasyActivity/.gitattributes
alexndlm: Thanks.

alexndlm (approved):

roman-eonx (approved):

natepage (approved):

PR #1075, opened 2022-11-21
Rename quality tools folder

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

In other projects, we don't use dot for the Docker folder or other tool-related folders.


roman-eonx:
> In other projects, we don't use dot for the Docker folder or other tool-related folders.

But we do use a dot in the `.github` folder name. And this is intentional, as it allows us to differentiate "technical" folders that are not required for the app/project/package code to work. We either should put a dot into `docker`, `bin`, etc., or remove it everywhere. Otherwise, this PR doesn't add any value.

What are the profits of the new naming? And was your plan to rename the folder in all the projects?

alexndlm:
Now we have dot only for `.github` and `.quality`.
`.github` - is it GitHub feature, and we can't change this for now.

An overall plan for other projects - to move `quality` from `src` to the project root folder.

roman-eonx:
Alright, the idea of moving `quality` to the root in all the projects looks good. Let's do it.

roman-eonx (approved):

natepage (approved):

PR #1085, opened 2023-01-11
Use SentStamp instead of ConsumedByWorkerStamp

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Use `SentStamp` to make the package work with Sync transport.

Symfony Messenger runs all middleware two times for each message.  We use `ConsumedByWorkerStamp` to determine if it is a first or second middleware dispatch.

`ConsumedByWorkerStamp` is set only in `symfony/messenger/Worker.php:158`. That's why EasyBatch doesn't work with Sync transport.

In this PR, we start using `SentStamp` for the purpose above. This stamp sets both in `Worker` and `SyncTransport`.

A small test app was added to clarify Symfony Functional tests setup.


ERuban:
Am I right that we need this only for testing purposes? Bcs we handling such things in the worker, so `ConsumedByWorkerStamp` looks nice. And as far as I remember we testing this thing in tests emulating worker. 

I'm a bit worried about using `SentStamp` instead of `ConsumedByWorkerStamp`...

akl-95:
@alexndlm Sorry, what do you mean by `This is visible in a \Symfony\Component\Messenger\Middleware\FailedMessageProcessingMiddleware - ReceivedStamp is not required.`?

There is `ReceivedStamp` usage in the `FailedMessageProcessingMiddleware`.

alexndlm:
We decided to continue to use async messages in the tests.
See https://github.com/eonx-com/easy-monorepo/pull/1088

PR #1088, opened 2023-01-24
[EasyTest] Improve Symfony Messenger asserts

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes

This PR provides small BC for tests, that can be easy fixed.

### DIFF
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -65,6 +65,12 @@ trait ExceptionTrait
         self::$isInsideSafeCall = false;
     }
 
+    /**
+     * This check is needed to make sure that the developer is not forgetting to assert the thrown exception,
+     * when using safeCall().
+     *
+     * This check should be called in the tearDown() method of the test class.
+     */
     private function checkThrownExceptionAsserted(): void
     {
         if ($this->isNeedToAssertThrownException) {
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -1,0 +1,76 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyTest\Traits;
+
+use Closure;
+use RuntimeException;
+use Throwable;
+
+trait ExceptionTrait
+{
+    protected static bool $isInsideSafeCall = false;
+
+    protected ?Throwable $thrownException = null;
+
+    private bool $isNeedToAssertThrownException = false;
+
+    /**
+     * @param class-string<\Throwable>|null $previousException
+     */
+    protected function assertThrownException(
+        string $expectedException,
+        ?int $code = null,
+        ?string $previousException = null
+    ): void {
+        $this->isNeedToAssertThrownException = false;
+
+        self::assertNotNull($this->thrownException);
+
+        if ($this->thrownException instanceof $expectedException === false) {
+            throw $this->thrownException;
+        }
+
+        if (\str_starts_with($expectedException, 'App\\')) {
+            $message = "Didn't you forget to define the error code?";
+            self::assertNotNull($code, $message);
+            self::assertNotSame(0, $code, $message);
+        }
+
+        if ($code !== null) {
+            self::assertSame($code, $this->thrownException->getCode());
+        }
+
+        if ($previousException === null) {
+            self::assertNull($this->thrownException->getPrevious());
+        }
+
+        if ($previousException !== null) {
+            self::assertInstanceOf($previousException, $this->thrownException->getPrevious());
+        }
+    }
+
+    protected function safeCall(Closure $func): void
+    {
+        self::$isInsideSafeCall = true;
+
+        try {
+            $this->isNeedToAssertThrownException = true;
+            $func();
+        } catch (Throwable $exception) {
+            $this->thrownException = $exception;
+        }
+
+        self::$isInsideSafeCall = false;
+    }
+
+    private function checkThrownExceptionAsserted(): void
+    {
+        if ($this->isNeedToAssertThrownException) {
+            throw new RuntimeException(
+                'ExceptionTrait::safeCall() must be followed by ExceptionTrait::assertThrownException()'
+            );
+        }
+    }
+}
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -8,6 +8,9 @@ use Closure;
 use RuntimeException;
 use Throwable;
 
+/**
+ * @mixin \PHPUnit\Framework\TestCase
+ */
 trait ExceptionTrait
 {
     protected static bool $isInsideSafeCall = false;
@@ -17,6 +20,7 @@ trait ExceptionTrait
     private bool $isThrownExceptionAssertionNeeded = false;
 
     /**
+     * @param class-string<\Throwable> $expectedException
      * @param class-string<\Throwable>|null $previousException
      */
     protected function assertThrownException(
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -14,7 +14,7 @@ trait ExceptionTrait
 
     protected ?Throwable $thrownException = null;
 
-    private bool $isNeedToAssertThrownException = false;
+    private bool $isThrownExceptionAssertionNeeded = false;
 
     /**
      * @param class-string<\Throwable>|null $previousException
@@ -24,7 +24,7 @@ trait ExceptionTrait
         ?int $code = null,
         ?string $previousException = null
     ): void {
-        $this->isNeedToAssertThrownException = false;
+        $this->isThrownExceptionAssertionNeeded = false;
 
         self::assertNotNull($this->thrownException);
 
@@ -56,7 +56,7 @@ trait ExceptionTrait
         self::$isInsideSafeCall = true;
 
         try {
-            $this->isNeedToAssertThrownException = true;
+            $this->isThrownExceptionAssertionNeeded = true;
             $func();
         } catch (Throwable $exception) {
             $this->thrownException = $exception;
@@ -73,7 +73,7 @@ trait ExceptionTrait
      */
     private function checkThrownExceptionAsserted(): void
     {
-        if ($this->isNeedToAssertThrownException) {
+        if ($this->isThrownExceptionAssertionNeeded) {
             throw new RuntimeException(
                 'ExceptionTrait::safeCall() must be followed by ExceptionTrait::assertThrownException()'
             );
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -71,7 +71,7 @@ trait ExceptionTrait
      *
      * This check should be called in the tearDown() method of the test class.
      */
-    private function checkThrownExceptionAsserted(): void
+    private function checkThrownExceptionAssertion(): void
     {
         if ($this->isThrownExceptionAssertionNeeded) {
             throw new RuntimeException(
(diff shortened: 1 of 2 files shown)


packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -51,45 +53,42 @@ public static function assertCountOfMessagesSentToTransport(
...
-     *
-     * @phpstan-param class-string $messageClass
+     * @param class-string $messageClass
+     * @param array<string, mixed> $expectedProperties
      */
     public static function assertMessageSentToFailedTransport(
         string $messageClass,
-        ?array $expectedProperties = null,
-        ?int $messagesCount = null
+        array $expectedProperties = [],
```
akl-95, line 75: Why do we need these default values? And why doesn't it break quality tools?

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -113,22 +112,31 @@ static function (object $message) use ($messageClass, $expectedProperties, $prop
...
     }
 
-    public static function consumeAsyncMessages(): void
+    /**
+     * @param array<int, array<class-string<\Throwable>, int|string>> $expectedExceptions
+     */
+    public static function consumeAsyncMessages(array $expectedExceptions = []): void
     {
-        self::runMessengerWorker('async', 'default');
+        if (isset(self::$isInsideSafeCall) && self::$isInsideSafeCall) {
```
akl-95, line 123: The `isInsideSafeCall` property is from another trait. We shoud avoid mixing traits as much as possible. Is this really required?

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -113,22 +112,31 @@ static function (object $message) use ($messageClass, $expectedProperties, $prop
...
+            throw new RuntimeException(
+                "You can't use MessengerAssertionsTrait::consumeAsyncMessages() in ExceptionTrait::safeCall()"
```
akl-95, line 125: Could you please comment why we can't?

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -206,8 +220,58 @@ private static function runMessengerWorker(string $transportName, string $busNam
             $messageLimitListener = new StopWorkerOnMessageLimitListener($messagesCount);
             $eventDispatcher->addSubscriber($messageLimitListener);
 
-            $worker = new Worker(['async' => $asyncTransport], $messageBus, $eventDispatcher);
+            $worker = new Worker(['async' => $transport], $messageBus, $eventDispatcher);
             $worker->run();
         }
+
+        if (\count((array)$transport->get()) > 0) {
+            throw new \RuntimeException('Unable to consume all messages from async transport.');
```
akl-95, line 228: Please remove `\`.

akl-95, line 228: What case does this cover?

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -206,8 +220,58 @@ private static function runMessengerWorker(string $transportName, string $busNam
...
+        }
+
+        foreach ($transport->getRejected() as $envelope) {
+            /** @var \Symfony\Component\Messenger\Stamp\ErrorDetailsStamp $errorDetailsStamp */
+            foreach ($envelope->all(ErrorDetailsStamp::class) as $errorDetailsStamp) {
+                foreach ($expectedExceptions as $key => $expectedExceptionPair) {
```
akl-95, line 234: Do we really need the expected exceptions feature? Is not exposing of occurred errors enough?

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -206,8 +220,58 @@ private static function runMessengerWorker(string $transportName, string $busNam
...
+                            === $errorDetailsStamp->getExceptionCode()
+                        )
+                    ) {
+                        unset($expectedExceptions[$key]);
+
+                        continue 2;
+                    }
+                }
+
+                echo "\n";
```
akl-95, line 248: What about private `printErrorDetails` method?

packages/EasyTest/src/Traits/ExceptionTrait.php
```diff
@@ -0,0 +1,76 @@
...
+        } catch (Throwable $exception) {
+            $this->thrownException = $exception;
+        }
+
+        self::$isInsideSafeCall = false;
+    }
+
+    private function checkThrownExceptionAsserted(): void
+    {
+        if ($this->isNeedToAssertThrownException) {
```
akl-95, line 70: If I understand correctly the `isNeedToAssertThrownException` is introduced for this check only. But why do we need to force `assertThrownException`? Do we have any issues with using `safeCall` now?

akl-95 (changes requested):

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: This is the final code, so default values are more useful.

This trait is not used in other packages, and PHPStan doesn't check it.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Yes, it is a useful feature. This will help avoid hiding bugs in tests.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Starting a worker should always work without exceptions.

If there are exceptions, we should not catch it.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: This is a case where one message triggers other messages, and those messages trigger more messages, and so on.

Here we check that we have processed all the messages in the test.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: We have tests where we expect that message will be failed.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Thanks. Fixed.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Done

packages/EasyTest/src/Traits/ExceptionTrait.php
alexndlm: Yes, we have some tests with redundant using `safeCall`.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
akl-95: What do you mean by final code? This method may be overrided on the app level.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
akl-95: What kind of bugs? Just to understand and capture this in the PR.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
akl-95: Let's suppose I expect an exception from the message handler and want to use `safeCall` for this. How should we handle this situation?

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
akl-95: OK, thanks. But what should we do then if some message hadler dispatched another message?

packages/EasyTest/src/Traits/ExceptionTrait.php
akl-95: Are you talking about the cases when we use `safeCall` but no exception is thrown in fact?

alexndlm: Yes.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: I mean that this code is not part of some abstract class or interface and will not be used for inheritance.

Using `null` values is useless.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Explanation added in doc comments.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Explanation added in doc comments.

Expected exceptions should be provided in the `$expectedExceptions` argument.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Explanation added in doc comments.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Explanation added in cod comments.

akl-95 (dismissed):

akl-95 (approved):

packages/EasyTest/src/Traits/ExceptionTrait.php
```diff
@@ -0,0 +1,82 @@
...
+     * This check should be called in the tearDown() method of the test class.
+     */
+    private function checkThrownExceptionAssertion(): void
+    {
+        if ($this->isThrownExceptionAssertionNeeded) {
+            throw new RuntimeException(
+                'ExceptionTrait::safeCall() must be followed by ExceptionTrait::assertThrownException()'
+            );
+        }
+    }
```
BOB41K1987, line 81: In PBA projects we have a [`getErrorHandler` method](https://github.com/eonx-com/mastercard-pba-transact-api-v1/blob/release/src/tests/Util/ExceptionTrait.php) and we use it like so. Don't we need it here as well? 
image

BOB41K1987 (changes requested):

packages/EasyTest/src/Traits/ExceptionTrait.php
```diff
@@ -0,0 +1,82 @@
...
+trait ExceptionTrait
+{
+    protected static bool $isInsideSafeCall = false;
+
+    protected ?Throwable $thrownException = null;
+
+    private bool $isThrownExceptionAssertionNeeded = false;
+
+    /**
+     * @param class-string<\Throwable>|null $previousException
```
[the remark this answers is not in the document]
alexndlm: Done.

packages/EasyTest/src/Traits/ExceptionTrait.php
```diff
@@ -0,0 +1,82 @@
...
+     */
+    protected function assertThrownException(
+        string $expectedException,
+        ?int $code = null,
+        ?string $previousException = null
```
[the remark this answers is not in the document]
alexndlm: We decided not to assert an exception message.

packages/EasyTest/src/Traits/ExceptionTrait.php
alexndlm: This will be implemented in another way - new stub with static methods.

akl-95 (approved):

BOB41K1987 (approved):

roman-eonx (approved):

natepage (approved):

PR #1091, opened 2023-02-09
[EasyErrorHandler] feature/easy-error-handler-enum-support

alexandrtspl:
* collect error codes from enums

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no 
| Tests pass?   | yes    

Find all enums with attribute `#[AsErrorCodes]` in the project and collect error codes.

Command `../scripts/console.sh easy-error-handler:error-codes:analyze` output example:
```
+--------------------------------------------+------------------------+
| Error code group                           | Next error code to use |
+--------------------------------------------+------------------------+
| ErrorActorNotResolved*                     | 12801                  |
| ErrorApiResource*                          | 10405                  |
| ErrorAuth*                                 | 10614                  |
| ErrorAuthorityAgreementDuplicateActive*    | 10501                  |
| ErrorBankAccountIsIneligibleFor*           | 13504                  |
| ErrorBankAccountMdes*                      | 13407                  |
| ErrorBasiqApi*                             | 12704                  |
| ErrorBpayApi*                              | 10704                  |
| ErrorCreditCard*                           | 10810                  |
| ErrorCuscal*                               | 13303                  |
.....skipped.....
| ErrorPayIdSenderIneligibleForDeactivation* | 11901                  |
| ErrorPaymentConfigTemplate*                | 13106                  |
| ErrorPaymentMethod*                        | 11803                  |
| ErrorPointsWebhook*                        | 12004                  |
| ErrorReconciliation*                       | 12118                  |
| ErrorSplit*                                | 12212                  |
| ErrorTransaction*                          | 12303                  |
| ErrorVerification*                         | 12408                  |
+--------------------------------------------+------------------------+

The error code for the new group is 13800.

```


packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+        return $errorCodes;
+    }
+
+    /**
+     * @return array<mixed>
+     */
+    private function locateErrorCodesEnums(): array
```
alexndlm, line 44: What about rewrite with `nikic/php-parser`.

Chat GPT suggest like this:
```php
<?php

use PhpParser\Error;
use PhpParser\ParserFactory;

$interfaceName = 'MyInterface';

$parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);

$directory = new RecursiveDirectoryIterator(__DIR__);
$iterator = new RecursiveIteratorIterator($directory);
$regex = new RegexIterator($iterator, '/\.php$/');

foreach ($regex as $file) {
    try {
        $code = file_get_contents($file->getPathname());
        $stmts = $parser->parse($code);

        foreach ($stmts as $stmt) {
            if ($stmt instanceof \PhpParser\Node\Stmt\Class_) {
                $implements = $stmt->implements;
                foreach ($implements as $impl) {
                    if ($impl->toString() === $interfaceName) {
                        echo "Class " . $stmt->name . " implements " . $impl->toString() . " in file " . $file->getPathname() . "\n";
                    }
                }
            }
        }
    } catch (Error $e) {
        // handle parse error
    }
}
```

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
```diff
@@ -0,0 +1,98 @@
...
+namespace EonX\EasyErrorHandler\Bridge\Symfony\Providers;
+
+use EonX\EasyErrorHandler\Interfaces\ErrorCodesEnumInterface;
+use EonX\EasyErrorHandler\Interfaces\ErrorCodesProviderInterface;
+use ReflectionClass;
+use Symfony\Component\Finder\Finder;
+
+final class ErrorCodesByEnumProvider implements ErrorCodesProviderInterface
+{
+    private const ERROR_CODE_NAME_PREFIX = 'Error';
```
alexndlm, line 14: I believe this is some redundant. We may use all enum items.

Wdyt?

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
```diff
@@ -60,10 +64,13 @@ public function process(): ErrorCodesDto
     /**
      * @param mixed[] $errorCodeNames
      */
-    private function determineCategoryName(array $errorCodeNames): string
+    private function determineCategoryName(array $errorCodeNames, bool $isEnum): string
     {
         $explodedErrorCodeNames = \array_map(
-            static function ($errorCodeName): array {
+            static function ($errorCodeName) use ($isEnum): array {
+                if ($isEnum === true) {
```
alexndlm, line 71: I believe this is not needed. We use capitalised enum case names, like `case SomeErrorName = 12345;`.

I believe `SomeErrorName` will be more readable.

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -91,12 +93,22 @@
         ->arg('$defaultIsVerbose', param(BridgeConstantsInterface::PARAM_IS_VERBOSE));
 
     // Error codes provider
-    $services->set(ErrorCodesProviderInterface::class, ErrorCodesProvider::class)
+    $services->set('error_codes_provider.by_interface', ErrorCodesProvider::class)
```
alexndlm, line 96: Can we rename `ErrorCodesProvider` to `ErrorCodesByInterfaceProvider` without BC?

alexndlm (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexandrtspl: I really need to install phpParser just because chatGPT suggest it?

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexandrtspl: I don't understand your comment. 
This code needed only to be split by groups later. Nothing happened with names here. 
It will be like in the example: `SomeErrorName`

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
alexandrtspl: Renamed

alexandrtspl: Done

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexandrtspl: Agreed, removed

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexandrtspl: Anyway done. Please check.

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexndlm: Thanks

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
alexndlm: Thanks

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
```diff
@@ -60,10 +64,13 @@ public function process(): ErrorCodesDto
...
+                    $errorCodeName = \preg_replace('/([a-z])([A-Z])/u', '$1_$2', $errorCodeName);
```
alexndlm, line 72: Let's move this variable inside `determineCategoryName`

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexndlm: We decided to continue with attributes.

alexndlm (changes requested):

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexandrtspl: Done

composer.json
```diff
@@ -219,6 +219,7 @@
     "config": {
         "process-timeout": 0,
         "allow-plugins": {
+            "php-http/discovery": false,
```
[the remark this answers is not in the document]
alexandrtspl: For another package. My bad. Removed.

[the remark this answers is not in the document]
alexandrtspl: php-http/discovery 1.15.1 - new feature - this plugin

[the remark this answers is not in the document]
alexndlm: Got it. Thanks

packages/EasyErrorHandler/composer.json
```diff
@@ -7,7 +7,8 @@
         "php": "^8.1",
         "eonx-com/easy-logging": "^4.4",
         "eonx-com/easy-utils": "^4.4",
-        "nesbot/carbon": "^2.22"
+        "nesbot/carbon": "^2.22",
+        "nikic/php-parser": "^4.15"
```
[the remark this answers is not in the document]
alexndlm: Got it. Thanks

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexndlm: Thanks

alexndlm (changes requested):

composer.json
alexndlm, line 222: This is not about this PR, but let's also fix EasyApiToken tests.

alexndlm (changes requested):

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
```diff
@@ -0,0 +1,85 @@
...
+
+        return $errorCodes;
+    }
+
+    /**
+     * @return array<mixed>
+     */
+    private function locateErrorCodesEnums()
+    {
+        $parser = (new ParserFactory())->create(ParserFactory::PREFER_PHP7);
```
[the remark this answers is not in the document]
alexndlm: Or even better, let's instantiate it in the constructor.

Symfony DI should instantiate service only if needed so we wouldn't have any memory leaks.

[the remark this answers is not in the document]
alexandrtspl: Done

composer.json
alexandrtspl: Done, plugin disabled in EasyApiToken too

alexndlm: Thanks

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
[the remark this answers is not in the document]
alexndlm: Thanks

alexndlm (dismissed):

packages/EasyErrorHandler/src/Interfaces/ErrorCodesEnumInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace EonX\EasyErrorHandler\Interfaces;
+
+interface ErrorCodesEnumInterface
```
akl-95, line 7: We have the `AsErrorCodes` attribute now, why do we need the interface, it's unused.

packages/EasyErrorHandler/docs/config.md
```diff
@@ -47,13 +47,14 @@ Laravel has the following additional configuration option:
...
-| `transform_validation_errors`      | `true`     | Transform validation errors response structure based on ApiPlatformValidationErrrorResponseBuilder.      |
-| `translation_domain`               | `messages` | Symfony translation domain.                                                                              |
-| `verbose`                          | `false`    | Use extended error response containing exception message, trace, etc.                                    |
+| Configuration                      | Default    | Description                                                                                                         |
+|------------------------------------|------------|---------------------------------------------------------------------------------------------------------------------|
+| `bugsnag_ignore_validation_errors` | `true`     | Ignore validation errors based on ApiPlatformValidationErrorsResponseBuilder                                        |
+| `error_codes_enums_auto_discovery` | `false`    | Try recursively to find all enums with error codes that implement ErrorCodesEnumInterface in the project directory. |
```
akl-95, line 53: Please fix `error_codes_enums...` -> `error_code_enums...` here and in other places.

akl-95, line 53: Please fix the description if we base our logic on the `AsErrorCodes` attribute like so `Find all error code enums having AsErrorCode attribute in the project directory recursively.`.

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/EasyErrorHandlerExtension.php
```diff
@@ -76,6 +78,12 @@ public function load(array $configs, ContainerBuilder $container): void
             BridgeConstantsInterface::PARAM_ERROR_CODES_CATEGORY_SIZE,
             $config['error_codes_category_size']
         );
+        $container->setParameter(
```
akl-95, line 81: Why do we need this logic in this class?

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * @return array<mixed>
+     */
+    private function locateErrorCodesEnums()
+    {
+        $directory = new \RecursiveDirectoryIterator($this->projectDir . '/src');
+        $iterator = new \RecursiveIteratorIterator($directory);
+        $regex = new \RegexIterator($iterator, '/Enum\/.*Enum\.php$/');
```
akl-95, line 51: What if there is another namespace structure in the app? For example, `Enums` folder instead of `Enum`. We have an exact criteria - the attribute, so why do we need to introduce the regex?

packages/EasyErrorHandler/docs/config.md
```diff
@@ -47,13 +47,14 @@ Laravel has the following additional configuration option:
 
 Symfony has the following additional configuration options:
 
-| Configuration                      | Default    | Description                                                                                              |
```
akl-95, line 50: I see we don't have `errorCodesInterface` in the common configuration section above. Let's add it.

packages/EasyErrorHandler/docs/config.md
akl-95: And do we really need this option? We use `AsErrorCodes` for the only purpose only - to allow EasyErrorHandler collect all the implementations. So if you use this attribute the package can automatically collect enums.

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
```diff
@@ -62,8 +65,12 @@ public function process(): ErrorCodesDto
      */
     private function determineCategoryName(array $errorCodeNames): string
     {
+        $isEnum = $this->errorCodesSource === ErrorCodesProviderLocatorInterface::SOURCE_ENUM;
```
akl-95, line 68: Until this processor knows about enums the `ErrorCodesProviderInterface` makes no sense because we still have to know how to work with different types of source. So ideally we should try to move the source-specific logic into the relevant provider and work with some high-level data only in the processor. Let's try to do it.

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
```diff
@@ -7,21 +7,24 @@
...
-use EonX\EasyErrorHandler\Interfaces\ErrorCodesProviderInterface;
+use EonX\EasyErrorHandler\Interfaces\ErrorCodesProviderLocatorInterface;
 
 final class ErrorCodesGroupProcessor implements ErrorCodesGroupProcessorInterface
 {
     public function __construct(
-        private int $categorySize,
-        private ErrorCodesProviderInterface $errorCodesProvider
+        private readonly int $categorySize,
+        private readonly string $errorCodesSource,
```
akl-95, line 16: At the first glance we make the configuration too complex. For example you can set source to enum and leave auto discovery = false. Or source = enum and set error codes interface. Both these options won't work and make no sense actually. The easiest way to me is to support both source from the start. We can introduce a set of providers and go through them in the processor. Each provider uses its own logic to fetch the codes: interface provider uses `error_codes_interface` configuration (and returns an empty result if the configuration is not set) and the enum provider uses all `AsErrorCodes` implementation without any explicit configuration. Wdyt?

akl-95 (changes requested):

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexandrtspl: You can't set the source manually it's set automatically in Configuration.php if errorCodeEnumsAutodiscovery = true.

alexandrtspl: But I understood complexity. I'll remove errorCodeEnumAutodiscovery and use it only if errorCodesInterface is not set.

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexandrtspl: We need it to avoid parsing all files in the project. It's prefilter. 

Maybe add a configuration parameter for this? And by default use `*.php`

Wdyt?

packages/EasyErrorHandler/src/Interfaces/ErrorCodesEnumInterface.php
alexandrtspl: My bad. Removed

packages/EasyErrorHandler/docs/config.md
alexandrtspl: Fixed

alexandrtspl: error_codes_enums_auto_discovery removed

packages/EasyErrorHandler/src/Bridge/Symfony/DependencyInjection/EasyErrorHandlerExtension.php
alexandrtspl: Not needed anymore. Removed

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexandrtspl: Fixed

packages/EasyErrorHandler/docs/config.md
alexandrtspl: Added. Please check it

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexandrtspl: Refactored. Please check it. Wdyt?

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexandrtspl: Configuration changed. `errorCodeEnumAutodiscovery` removed. Now it's used only if errorCodesByInterfaceProvider returned empty array

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
```diff
@@ -91,12 +91,19 @@
...
         ->arg('$errorCodesInterface', param(BridgeConstantsInterface::PARAM_ERROR_CODES_INTERFACE));
 
+    $services->set('error_codes_provider.by_enum', ErrorCodesByEnumProvider::class)
+        ->arg('$projectDir', param('kernel.project_dir'));
+
     // Error codes group processor
     $services->set(ErrorCodesGroupProcessorInterface::class, ErrorCodesGroupProcessor::class)
-        ->arg('$categorySize', param(BridgeConstantsInterface::PARAM_ERROR_CODES_CATEGORY_SIZE));
+        ->arg('$categorySize', param(BridgeConstantsInterface::PARAM_ERROR_CODES_CATEGORY_SIZE))
+        ->arg('$errorCodesProviders', [
```
alexndlm, line 103: Let's use a tagged iterator.

alexndlm (changes requested):

alexndlm (changes requested):
Let's add to `\EonX\EasyErrorHandler\Exceptions\BaseException`:
```php
public function __construct(string $message = '', BackedEnum|int $code = 0, ?Throwable $previous = null)
{
    parent::__construct($message, $code instanceof BackedEnum ? (int)$code->value : $code, $previous);
}
```

packages/EasyErrorHandler/src/Bridge/Symfony/Resources/config/services.php
alexandrtspl: Done

alexndlm: Thanks

alexndlm (approved):

packages/EasyErrorHandler/docs/config.md
```diff
@@ -17,27 +17,27 @@ called `easy_error_handler`.
```
akl-95, line 17: Let's add an info about how the package works with different error code sources.

packages/EasyErrorHandler/src/Providers/ErrorCodesByInterfaceProvider.php
```diff
@@ -9,7 +9,7 @@
 use ReflectionException;
 use Symfony\Component\ErrorHandler\Error\ClassNotFoundError;
 
-final class ErrorCodesProvider implements ErrorCodesProviderInterface
+final class ErrorCodesByInterfaceProvider implements ErrorCodesProviderInterface
```
akl-95, line 12: What about `ErrorCodesByInterfaceProvider` -> `ErrorCodesFromInterfaceProvider` and `ErrorCodesFromEnumsProvider` respectively? And the same for all aliases in configuration.

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
```diff
@@ -0,0 +1,88 @@
...
+use EonX\EasyErrorHandler\Annotations\AsErrorCodes;
+use EonX\EasyErrorHandler\Interfaces\ErrorCodesProviderInterface;
+use PhpParser\Error;
+use PhpParser\Node\Stmt\Enum_;
+use PhpParser\Node\Stmt\Namespace_;
+use PhpParser\Parser;
+use PhpParser\ParserFactory;
+use ReflectionClass;
+
+final class ErrorCodesByEnumProvider implements ErrorCodesProviderInterface
```
akl-95, line 16: Let's move it from the Symfony bridge to the common level, we can easily reuse it for Laravel as well.

akl-95 (changes requested):

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
akl-95: Let's try to make it responsibility of providers.

alexandrtspl: Refactored. Please check it

packages/EasyErrorHandler/docs/config.md
alexandrtspl: Added

packages/EasyErrorHandler/src/Providers/ErrorCodesByInterfaceProvider.php
alexandrtspl: Renamed

packages/EasyErrorHandler/src/Bridge/Symfony/Providers/ErrorCodesByEnumProvider.php
alexandrtspl: ErrorCodesFromEnumsProvider moved to common folder

packages/EasyErrorHandler/src/Interfaces/ErrorCodesEnumInterface.php
akl-95, line 7: Please remove, it's unused.

akl-95 (changes requested):

packages/EasyErrorHandler/docs/error-codes-analyzing.md
```diff
@@ -0,0 +1,68 @@
...
+### Error codes from enums
+
+`ErrorCodesFromEnumsProvider` expects error codes in `PascalCase` naming convention.
+
+Example:
+```
+#[AsErrorCodes]
+enum ErrorCodesEnum: int {
+
+    case ErrorSomeErrorOne = 100;
```
roman-eonx, line 51: Are we required to have this `Error` prefix? It looks redundant when you read code with some enum value.

packages/EasyErrorHandler/src/DataTransferObjects/ErrorCodeDto.php
```diff
@@ -0,0 +1,42 @@
...
+        private int $errorCode,
+        private array $splittedName,
+        private ?string $groupSeparator = null
+    ) {
+    }
+
+    /**
+     * @return array<int, string>
+     */
+    public function getSplittedName(): array
```
roman-eonx, line 23: Please change `Splitted` to `Split`, it's a more correct form.

packages/EasyErrorHandler/src/Providers/ErrorCodesFromEnumProvider.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return $errorCodes;
+    }
+
+    /**
+     * @return array<mixed>
+     */
+    public function locateErrorCodesEnums()
+    {
+        $directory = new \RecursiveDirectoryIterator($this->projectDir . '/src');
```
roman-eonx, line 61: In Laravel projects, we keep all the application code in the `app` directory. If we want to support Laravel projects, we should probably inject the `src`/`app` directory in the provider.

roman-eonx (changes requested):

packages/EasyErrorHandler/docs/error-codes-analyzing.md
akl-95: Yep, we got rid of this prefix in PG PR. @alexandrtspl let's remove it here too.

packages/EasyErrorHandler/src/Providers/ErrorCodesFromEnumProvider.php
alexandrtspl: My bad. Fixed

packages/EasyErrorHandler/src/DataTransferObjects/ErrorCodeDto.php
alexandrtspl: Done

packages/EasyErrorHandler/src/Interfaces/ErrorCodesEnumInterface.php
alexandrtspl: Done

akl-95 (approved):

alexndlm (approved):

roman-eonx (approved):

packages/EasyErrorHandler/src/Bridge/Laravel/Provider/EasyErrorHandlerServiceProvider.php
```diff
@@ -197,18 +198,27 @@ static function (Container $app): BugsnagReporterProvider {
...
-                return new ErrorCodesProvider(\config('easy-error-handler.error_codes_interface'));
+                return new ErrorCodesFromInterfaceProvider(\config('easy-error-handler.error_codes_interface'));
             }
         );
+        $this->app->tag(ErrorCodesFromInterfaceProvider::class, [BridgeConstantsInterface::TAG_ERROR_CODES_PROVIDER]);
+
+        $this->app->singleton(
+            ErrorCodesFromEnumProvider::class,
+            function (): ErrorCodesProviderInterface {
+                return new ErrorCodesFromEnumProvider($this->app->basePath() . '/app');
```
natepage, line 211: Instead of using `$this->app` please use a static function and expect an argument `Container $app`.

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
```diff
@@ -7,24 +7,34 @@
...
 final class ErrorCodesGroupProcessor implements ErrorCodesGroupProcessorInterface
 {
+    /**
+     * @param array<\EonX\EasyErrorHandler\Interfaces\ErrorCodesProviderInterface> $errorCodesProviders
+     */
     public function __construct(
-        private int $categorySize,
-        private ErrorCodesProviderInterface $errorCodesProvider
+        private readonly int $categorySize,
+        private readonly iterable $errorCodesProviders
```
natepage, line 18: I would recommend to convert this iterable to a real array, because technically it will be a Generator and the code will be able to iterate through it only once so any attempt to list error codes with the same application instance will break.

packages/EasyErrorHandler/src/Providers/ErrorCodesFromEnumProvider.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return $errorCodes;
+    }
+
+    /**
+     * @return array<mixed>
+     */
+    public function locateErrorCodesEnums()
+    {
+        $directory = new \RecursiveDirectoryIterator($this->projectDir);
```
natepage, line 61: Maybe we could have used symfony finder here? 😄

natepage (changes requested):

packages/EasyErrorHandler/docs/error-codes-analyzing.md
alexandrtspl: Done

packages/EasyErrorHandler/src/Bridge/Laravel/Provider/EasyErrorHandlerServiceProvider.php
alexandrtspl: `Container` interface is not enough for `phpstan`. `Application $app` added. Wdyt?

packages/EasyErrorHandler/src/Processors/ErrorCodesGroupProcessor.php
alexandrtspl: Tagged iterator replaced with array

packages/EasyErrorHandler/src/Providers/ErrorCodesFromEnumProvider.php
alexandrtspl: Done.

natepage (approved):

PR #1094, opened 2023-02-18
[EasyAPiToken] Update firebase/php-jwt

alexndlm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Due to https://github.com/advisories/GHSA-8xf4-w7qw-pjjw

See https://github.com/firebase/php-jwt/releases/tag/v6.0.0


natepage (changes requested):
There is a few BC breaks in this PR, can we please try not to introduce any and support both php-jwt v5 and v6, otherwise we will have to wait for the next major release to get that in

packages/EasyApiToken/src/External/Auth0/TokenGenerator.php
```diff
@@ -76,6 +76,6 @@ public function generate(
             true
         ) : $this->secret;
 
-        return JWT::encode($payload, (string)$secret);
+        return JWT::encode($payload, (string)$secret, 'HS256');
```
itorgov, line 79: Please use the `EonX\EasyApiToken\Interfaces\AlgorithmsInterface::HS256` constant.

packages/EasyApiToken/src/External/AwsCognito/JwkFetcher.php
```diff
@@ -93,7 +74,7 @@ private function fetchKeysFromAws(UserPoolConfigInterface $userPoolConfig): arra
         $keys = [];
 
         foreach ($response['keys'] as $jwk) {
-            $keys[$jwk['kid']] = $this->convertJwkToPem($jwk);
+            $keys[(string)$jwk['kid']] = new Key($this->convertJwkToPem($jwk), $jwk['alg'] ?? 'HS256');
```
itorgov, line 77: The same here, the `\EonX\EasyApiToken\Interfaces\AlgorithmsInterface::HS256` const please.

itorgov (changes requested):

itorgov:
> There is a few BC breaks in this PR, can we please try not to introduce any and support both php-jwt v5 and v6, otherwise we will have to wait for the next major release to get that in



> There is a few BC breaks in this PR, can we please try not to introduce any and support both php-jwt v5 and v6, otherwise we will have to wait for the next major release to get that in

Agree 🙂

packages/EasyApiToken/src/External/AwsCognito/JwkFetcher.php
alexndlm: Refactored.

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
```diff
@@ -11,30 +11,33 @@
...
      */
     public function __construct(
-        private UserPoolConfigInterface $userPoolConfig,
-        ?JwkFetcherInterface $jwkFetcher = null,
+        private readonly UserPoolConfigInterface $userPoolConfig,
+        private JwkFetcherInterface $jwkFetcher = new JwkFetcher(),
         ?array $allowedAlgos = null,
-        private ?int $leeway = null
+        private readonly ?int $leeway = null,
+        private readonly string $defaultJwkAlgo = 'HS256'
```
ERuban, line 39: Shouldn't it be `RS256` by default for AwsCognitoJwtDriver? As I can see, we using `RS256` for Cognito JWT.
And, please, use the constant if possible

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
```diff
@@ -48,8 +51,16 @@ public function decode(string $token): object
             JWT::$leeway = $this->leeway;
         }
 
-        /** @var \stdClass $decodedToken */
-        $decodedToken = JWT::decode($token, $this->jwkFetcher->getJwks($this->userPoolConfig), $this->allowedAlgos);
+        $jwks = $this->jwkFetcher->getJwks($this->userPoolConfig);
+
+        if (self::isFirebaseJwtV5() === false) {
+            foreach ($jwks as $keyId => $key) {
+                $jwks[$keyId] = new Key($key, $this->defaultJwkAlgo ?? 'HS256');
```
ERuban, line 58: `$this->defaultJwkAlgo` is not nullable

packages/EasyApiToken/src/External/JwtTrait.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    private static function isFirebaseJwtV5(): bool
+    {
+        if (isset(self::$isFirebaseJwtV5) === false) {
+            $methodParams = \array_map(
+                static fn ($parameter) => $parameter->name,
+                (new ReflectionMethod(BaseJWT::class, 'decode'))->getParameters()
+            );
+
+            self::$isFirebaseJwtV5 = \in_array('allowed_algs', $methodParams, true);
```
ERuban, line 30: `allowed_algs` -> `allowed_algos`?

ERuban (changes requested):

packages/EasyApiToken/src/External/JwtTrait.php
alexndlm: Here ok - https://github.com/firebase/php-jwt/blob/83b609028194aa042ea33b5af2d41a7427de80e6/src/JWT.php#L83

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
alexndlm: Yes, you are right. Refactored.

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
alexndlm: Thanks. Fixed.

packages/EasyApiToken/src/External/Auth0/TokenGenerator.php
```diff
@@ -9,30 +9,18 @@
...
-     *
-     * @var string|null
-     */
-    private $audience;
-
-    /**
-     * @var null|string
-     */
-    private $issuer;
+    private const DEFAULT_ALG = 'HS256';
```
ERuban, line 12: `DEFAULT_ALGO`? :)

ERuban (changes requested):

packages/EasyApiToken/src/External/Auth0/TokenGenerator.php
alexndlm: Yes. Thanks

packages/EasyApiToken/src/External/JwtTrait.php
```diff
@@ -0,0 +1,35 @@
...
+namespace EonX\EasyApiToken\External;
+
+use Firebase\JWT\JWT as BaseJWT;
+use ReflectionMethod;
+
+/**
+ * @deprecated compatibility layer, will be removed in 5.0
+ * @internal
+ */
+trait JwtTrait
```
ERuban, line 14: Let's name it a bit more obviously?
Dunno, maybe smth like `FirebaseJwtVersionControlTrait`, or just `FirebaseJwtVersionTrait`?
wdyt?

packages/EasyApiToken/src/External/AwsCognito/JwkFetcher.php
```diff
@@ -17,33 +17,13 @@
 
 final class JwkFetcher implements JwkFetcherInterface
 {
-    /**
-     * @var int
-     */
-    private const CACHE_EXPIRY = 3600;
```
itorgov, line 23: Why did you delete it? Let's just rename to `DEFAULT_CACHE_EXPIRY`.

packages/EasyApiToken/src/External/JwtTrait.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * @deprecated compatibility layer, will be removed in 5.0
+     * @internal
+     */
+    private static function isFirebaseJwtV5(): bool
+    {
+        if (isset(self::$isFirebaseJwtV5) === false) {
+            $methodParams = \array_map(
+                static fn ($parameter) => $parameter->name,
+                (new ReflectionMethod(BaseJWT::class, 'decode'))->getParameters()
```
ERuban, line 27: Hm, maybe it would be better to use smth like `Composer\InstalledVersions::getVersion($packageName)`?
I don't like to have reflection usage on each request.
wdyt?

ERuban (changes requested):

itorgov (changes requested):

packages/EasyApiToken/src/External/JwtTrait.php
alexndlm: Then we will need extra packages.
@natepage do you mind?

Снимок экрана 2023-02-20 в 16 21 16

Снимок экрана 2023-02-20 в 16 21 03

ERuban: We don't need these packages. You can use `version_compare` PHP function to compare version you got from `Composer\InstalledVersions::getVersion($packageName)`.

packages/EasyApiToken/src/External/JwtTrait.php
akl-95: I like `FirebaseJwtVersionTrait`.

packages/EasyApiToken/src/External/JwtTrait.php
alexndlm: At least we need `composer-runtime-api` for `Composer\InstalledVersions::getVersion($packageName)`

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
```diff
@@ -48,8 +56,16 @@ public function decode(string $token): object
             JWT::$leeway = $this->leeway;
         }
 
-        /** @var \stdClass $decodedToken */
-        $decodedToken = JWT::decode($token, $this->jwkFetcher->getJwks($this->userPoolConfig), $this->allowedAlgos);
+        $jwks = $this->jwkFetcher->getJwks($this->userPoolConfig);
+
+        if (self::isFirebaseJwtV5() === false) {
```
akl-95, line 61: This check looks like we are OK with versions <= 4 in this `if`. Let's either do `isFirebaseJwtV5` -> `getFirebaseJwtVersion` or `isFirebaseJwtV5` -> `isFirebaseJwtVersionGt5`. Or anything like that. But I don't like the 2nd option actually, the name looks strange to me.

akl-95 (changes requested):

packages/EasyApiToken/src/External/JwtTrait.php
ERuban: But, looking to its description: 
```
composer-runtime-api 2.0 introduced a new Composer\InstalledVersions class
```
and looking to ours `vendor/composer` folder - we already have it.

natepage: Usually in such situation I try to identify an easy thing that differentiate 2 versions of the same package, a new class, a missing class, method, etc. So it's quite easy to understand which version of the package you're dealing with by checking if the class exists for example.

alexndlm: @ERuban, beyond belief, it works.

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
alexndlm: Refactored, please check.

packages/EasyApiToken/src/External/JwtTrait.php
alexndlm: Done

itorgov (approved):

ERuban (dismissed):

akl-95 (dismissed):

roman-eonx (approved):

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
```diff
@@ -48,8 +56,16 @@ public function decode(string $token): object
             JWT::$leeway = $this->leeway;
         }
 
-        /** @var \stdClass $decodedToken */
-        $decodedToken = JWT::decode($token, $this->jwkFetcher->getJwks($this->userPoolConfig), $this->allowedAlgos);
+        $jwks = $this->jwkFetcher->getJwks($this->userPoolConfig);
+
+        if (self::isFirebaseJwtV6()) {
+            foreach ($jwks as $keyId => $key) {
+                $jwks[$keyId] = new Key($key, $this->defaultJwkAlgo);
```
natepage, line 63: Are the JWKs already indexed by their keyId from the jwkFetcher? If not then it means that we have an array with each key twice in it, in raw format and as Key object.

packages/EasyApiToken/tests/AbstractFirebaseJwtTokenTestCase.php
```diff
@@ -43,7 +43,7 @@ protected function createFirebaseJwtDriver(
         OpenSSLAsymmetricKey|string|null $publicKey = null,
         OpenSSLAsymmetricKey|string|null $privateKey = null,
         ?array $allowedAlgos = null,
-        ?int $leeway = null
+        ?int $leeway = null,
```
natepage, line 46: We never had trailing comas on function arguments before

natepage (changes requested):

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
alexndlm: `$jwks` is `array<int|string, string>`. How can we have duplication?

packages/EasyApiToken/tests/AbstractFirebaseJwtTokenTestCase.php
alexndlm: Thanks. Fixed

packages/EasyApiToken/src/External/AwsCognitoJwtDriver.php
natepage: Because if the original `$jwks` isn't indexed by the keyIds and just by auto-increment integers then we will end up with having the original keys with auto-increments and the key objects indexed by their keyId

alexndlm: `\EonX\EasyApiToken\External\AwsCognito\JwkFetcher::fetchKeysFromAws` returns an array of keys, with `kid` as an index.

I can't get what I do wrong.

natepage: ok if that's the case then you're not doing anything wrong, I was asking for this confirmation

natepage (approved):

PR #1098, opened 2023-03-21
[EasyTest] Make exception error code optional

alexandrtspl:
* error code is optional

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes    

Exception error code is optional. 

Two options to use in test:
Current:
`self::consumeAsyncMessages([[FilesystemException::class => 0]]);`
New one:
`self::consumeAsyncMessages([FilesystemException::class]);`

And we can check exceptions without code: 
`$this->assertThrownException(FileParsingException::class);`

### DIFF
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -36,6 +36,14 @@ trait ExceptionTrait
             throw $this->thrownException;
         }
 
+        if (
+            $code === null
+            && $this->thrownException->getCode() !== 0
+            && \str_starts_with($expectedException, 'App\\')
+        ) {
+            self::assertNotNull($code, "Didn't you forget to define the error code?");
+        }
+
         if ($code !== null) {
             self::assertSame($code, $this->thrownException->getCode());
         }
--- a/packages/EasyTest/src/Traits/ExceptionTrait.php
+++ b/packages/EasyTest/src/Traits/ExceptionTrait.php
@@ -36,12 +36,6 @@ trait ExceptionTrait
             throw $this->thrownException;
         }
 
-        if (\str_starts_with($expectedException, 'App\\')) {
-            $message = "Didn't you forget to define the error code?";
-            self::assertNotNull($code, $message);
-            self::assertNotSame(0, $code, $message);
-        }
-
         if ($code !== null) {
             self::assertSame($code, $this->thrownException->getCode());
         }
--- a/packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
+++ b/packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
@@ -211,7 +211,7 @@ trait MessengerAssertionsTrait
     }
 
     /**
-     * @param array<int, class-string, array<class-string<\Throwable>, int|string>> $expectedExceptions
+     * @param array<int, class-string<\Throwable>|array<class-string<\Throwable>, int|string>> $expectedExceptions
      */
     private static function runMessengerWorker(
         string $transportName,
--- a/packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
+++ b/packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
@@ -118,7 +118,7 @@ trait MessengerAssertionsTrait
     }
 
     /**
-     * @param array<int, class-string, array<class-string<\Throwable>, int|string>> $expectedExceptions
+     * @param array<int, class-string<\Throwable>|array<class-string<\Throwable>, int|string>> $expectedExceptions
      */
     public static function consumeAsyncMessages(array $expectedExceptions = []): void
     {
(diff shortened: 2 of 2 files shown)


akl-95 (approved):

packages/EasyTest/src/Traits/ExceptionTrait.php
```diff
@@ -36,12 +36,6 @@ protected function assertThrownException(
             throw $this->thrownException;
         }
 
-        if (\str_starts_with($expectedException, 'App\\')) {
```
alexndlm, line 39: Wdyt about the following:
```php
        if (
            $code === null
            && $this->thrownException->getCode() !== 0
            && \str_starts_with($expectedException, 'App\\')
        ) {
            self::assertNotNull($code, "Didn't you forget to define the error code?");
        }
```

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -118,7 +118,7 @@ static function (object $message) use ($messageClass, $expectedProperties, $prop
     }
 
     /**
-     * @param array<int, array<class-string<\Throwable>, int|string>> $expectedExceptions
+     * @param array<int, class-string, array<class-string<\Throwable>, int|string>> $expectedExceptions
```
alexndlm, line 121: It looks like a mistake. It should be:
`@param array<int, class-string<\Throwable>|array<class-string<\Throwable>, int|string>> $expectedExceptions`

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
```diff
@@ -266,11 +266,20 @@ private static function runMessengerWorker(
         foreach ($transport->getRejected() as $envelope) {
             /** @var \Symfony\Component\Messenger\Stamp\ErrorDetailsStamp $errorDetailsStamp */
             foreach ($envelope->all(ErrorDetailsStamp::class) as $errorDetailsStamp) {
-                foreach ($expectedExceptions as $key => $expectedExceptionPair) {
+                foreach ($expectedExceptions as $key => $expectedExceptionPairOrClass) {
+                    // if $expectedExceptionPairOrClass is a string = exceptionClass
+                    if ($expectedExceptionPairOrClass === $errorDetailsStamp->getExceptionClass() &&
```
alexndlm, line 271: Do we need to add `\is_string($expectedExceptionPairOrClass)`?

alexndlm, line 271: It looks like we can combine these two `if` into a single `if`. Wdyt?

alexndlm (changes requested):

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexandrtspl: Fixed

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexandrtspl: No, strict comparison already do it

alexandrtspl: Sure. But it's split for readable purposes. It's more clear.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Looks the same.

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: I got it. Thanks.

alexndlm: I got it.

alexndlm (changes requested):

packages/EasyTest/src/Traits/MessengerAssertionsTrait.php
alexndlm: Thanks

alexndlm (approved):
Good job. Thanks

akl-95 (approved):

packages/EasyTest/src/Traits/ExceptionTrait.php
```diff
@@ -36,10 +36,12 @@ protected function assertThrownException(
...
-            $message = "Didn't you forget to define the error code?";
-            self::assertNotNull($code, $message);
-            self::assertNotSame(0, $code, $message);
+        if (
+            $code === null
```
natepage, line 40: How is this supposed to work? Because here we are checking if `$code` is null but then on line 44 we are asserting that `$code` is not null.

alexandrtspl: Yes, it's not clear but we expect "always true" assert here. Just because we check all that we need in the "if" block. We need to check user provides an error code if `thrownException->getCode() != 0` otherwise shows a message.

natepage (approved):

PR #1276, opened 2023-09-18
[EasyAsync] Close persistent db connections

natepage:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| BC breaks?    | no     
| Deprecations? | no 
| Tests pass?   | yes


packages/EasyAsync/src/Bridge/Symfony/Resources/config/definition.php
```diff
@@ -7,6 +7,17 @@
     $definition->rootNode()
         ->children()
             ->booleanNode('messenger_middleware_auto_register')->defaultTrue()->end()
+            ->arrayNode('doctrine')
```
itorgov, line 10: What about to put it under the `messenger_worker` section? This new feature is related to Messenger to and not only to Doctrine. More than that, the service container config file named as `messenger_doctrine_persistent_connections`.

itorgov: And if we will go this way then we also will need to rename the `PARAM_DOCTRINE_PERSISTENT_CONNECTIONS_MAX_IDLE_TIME` const a little bit.

itorgov (changes requested):

packages/EasyAsync/src/Bridge/Symfony/Resources/config/definition.php
natepage: @itorgov 😄 funny I actually asked myself the same question haha
But then I've told myself that the package is "easy-async" and this is the Symfony bridge, so technically everything is about Messenger and its worker haha

That's why I've put it there, wdyt?

itorgov: Yeah, I don't mind, but let's rename `messenger_doctrine_persistent_connections.php` to `doctrine_persistent_connections.php` then, please.

itorgov (approved):

PR #1644, opened 2025-04-03
[EasyServerless] Fix SecretsHelper implementation

natepage:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | yes     
| Deprecations? | yes 
| Tests pass?   | yes    

Yes this PR contains BC break but nobody is using this package for real yet, so it's kinda experimental and the previous implementation was not working as expected, so let's just fix it now 😄


roman-eonx (approved):

roman-eonx:
Let's simply treat the helper as internal, so there is no BC 😉

